PR #82, opened 2022-05-12
Shadow dom compatibility

### LINKED ISSUE #81: Shadow root compatibilty
# 🚀 Feature request

Able to still position elements correctly when the component is inside a shadow dom.

Document referenced in useSetup Is always the outer html document which bypasses the shadow root when the component is inside there. The base doc should be the shadow root for the positioning to work correctly.

## ✅ Proposed solution or API

Change all occurrences of `ownerDocument` to `getRootNode()`. This will return the regular html doc when not inside a shadow dom and the shadow root when the component is there. This makes all the positioning for tooltips/popovers work.

## 📝 Additional Information

https://developer.mozilla.org/en-US/docs/Web/API/Node/getRootNode

xylophonehero:
Closes #81 

## 📝 Description

Uses `getRootNode()` over `ownerDocument` for compatibility with shadow root

The browser api for fetching the root document

No

## 📝 Additional Information

There are further places where `ownerDocument` is used (createPortal...). I will also take a look at those

### DIFF
--- a/.changeset/mean-feet-raise.md
+++ b/.changeset/mean-feet-raise.md
@@ -1,0 +1,7 @@
+---
+"@zag-js/react": patch
+"@zag-js/solid": patch
+"@zag-js/vue": patch
+---
+
+Prefer `getRootNode` over `ownerDocument` to support shadow dom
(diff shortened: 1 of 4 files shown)


segunadebayo:
Thanks for working on this @xylophonehero. I'm really cautious about this as it breaks areas where we need something like `doc.body` or even `doc.createElement`.

Deciding in the code when to use the shadow root vs owner document is something we need to carefully consider as well.

xylophonehero:
Good points to think about. Only tried with the Tooltip so far but I'm sure portals will cause headaches. I will update you with my findings as they come

segunadebayo:
Closing in favor of #89.

Thank you for inspiring this idea 💖

xylophonehero:
Looks very nice.

Thank you for adding this feature in! 🥇

PR #92, opened 2022-05-15
fix: prefer fraction-digits over precision

### LINKED ISSUE #79: [number-input] precision 0 meaning/behavior
# 🐛 Bug report

Setting precision to 0 seems to disable rounding (that is, leave the value alone), as opposed to rounding the values to whole numbers (integers). Is this intentional behavior or a bug? If intentional, is there a way to round to whole numbers?

## 💥 Steps to reproduce

1. Go to https://zagjs.com/components/react/number-input
2. Change precision to "0" (if not already)
3. Type "12.34" in the "Enter number" input on the left
4. Tab out
5. Observe the value remains 12.34 as opposed to 12

## 💻 Link to reproduction

N/A, see Zag's own Number Input page linked above.

## 🧐 Expected behavior

I would expect it to round to the nearest whole number when precision is 0 (perhaps some other value like `undefined` would disable rounding all together?).

## 🧭 Possible Solution

Not sure.

## 🌍 System information

| Chakra UI Core   |     N/A      |
| Browser          |      Chrome, Firefox      |
| Operating System |   Windows 10, macOS      |

## 📝 Additional information

Thank you.

segunadebayo:
Closes #79

## 📝 Description

Resolve inconsistencies in the meaning of "precision" by splitting it into "minFractionDigits" and "maxFractionDigits"

### DIFF
--- a/.xstate/accordion.js
+++ b/.xstate/accordion.js
@@ -1,85 +1,85 @@
-"use strict"
+"use strict";
 
-var _xstate = require("xstate")
+var _xstate = require("xstate");
 
-const { actions, createMachine, assign } = _xstate
-const { choose } = actions
-const fetchMachine = createMachine(
-  {
-    id: "accordion",
-    initial: "unknown",
-    context: {
-      "isExpanded && canToggle": false,
-      "!isExpanded": false,
-    },
-    on: {
-      SET_VALUE: {
-        actions: "setValue",
-      },
-      UPDATE_CONTEXT: {
-        actions: "updateContext",
-      },
-    },
-    states: {
-      unknown: {
-        on: {
-          SETUP: {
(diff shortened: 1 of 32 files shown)


wKovacs64:
Thanks @segunadebayo, this looks great. Did `@zag-js/number-utils` somehow escape being published to the `dev` tag? I'm trying out these new fraction decimal options, but `@zag-js/number-input@dev` still depends on `@zag-js/number-utils@0.1.1` which doesn't include `FormatDecimalOptions` yet (so TS is upset). No biggie, just curious.

segunadebayo:
@wKovacs64, that was my bad. I've just pushed a new `dev` tag that includes the correct version of number-utils

PR #96, opened 2022-05-19
Transform origin for poppers

xylophonehero:
## 📝 Description

Transform origins will want to come from the arrow especially the case where the popper is slightly scaling into view. 
In the case that a popper is positioned at the start or end, we can use the dimensions of the trigger to calculate where this will be

Transform origins for start and end placement variants are just at the start and end of the floating box

## 🚀 New behavior

Transform origins are now calculated so they are placed in the centre of the trigger

No

### DIFF
--- a/packages/utilities/popper/src/middleware.ts
+++ b/packages/utilities/popper/src/middleware.ts
@@ -18,25 +18,28 @@ export const cssVars = {
  * Transform Origin Middleware
  * -----------------------------------------------------------------------------*/
 
-const transforms = {
-  top: "bottom center",
-  "top-start": "bottom left",
-  "top-end": "bottom right",
-  bottom: "top center",
-  "bottom-start": "top left",
-  "bottom-end": "top right",
-  left: "right center",
-  "left-start": "right top",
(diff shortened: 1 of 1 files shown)


segunadebayo:
Hi @xylophonehero,

Could show a demo of what a before and after looks like?

xylophonehero:
Before
https://user-images.githubusercontent.com/68996622/169564619-b01be5f7-0874-4a18-9d77-2b14985dc838.mov

After
https://user-images.githubusercontent.com/68996622/169564708-d3f2eac0-1c1a-41e4-8ac9-b4448be64e1d.mov

xylophonehero:
I did just notice that if the tooltip is small that the trigger then it doesn't work as well. If we get access to the arrow placement that would probably work better.

This was done in a private repo so can't share the code but I can put together a codepe next week

segunadebayo:
Thanks for the video @xylophonehero. Closing in favor of #100.

I managed to figure out how to use data from the arrow placement to compute the transform origin.

xylophonehero:
Nice looks good!

PR #113, opened 2022-05-26
chore: add `data-indeterminate` attribute to prop types

anubra266:
add `data-indeterminate` attribute to prop types


segunadebayo:
Please include this in the PR you're working on since there's where it's needed.

Thanks

PR #114, opened 2022-05-27
feat: add ability to set default dialog open state

alexluong:
Closes # 

## 📝 Description

Enable users to set whether dialog should open or close on default

Currently, all dialog is closed on default.

## 🚀 New behavior

If user pass `defaultIsOpen: true` to context, the dialog will transition to `open` state after setup.

No

## 📝 Additional Information

---

I also added a new Next.js page and Playwright test to test this behaviour.

I'd love to hear your input on these:

- config name: `defaultIsOpen`, I take this from Chakra UI naming convention, feel free to suggest alternatives
- the machine code / guard usage; as the guard is fairly straightforward, I just wrote the guard inline instead of giving it a name
- The name of the new route: "Dialog Default Open" - "/dialog-default-open"
- The e2e test to check if the dialog is open. I just check if the close button is focused.

If everything looks good, I'll follow up with the pages for Vue and Svelte as well.

And lastly, as this is my first PR, I'd appreciate some help with the next step (changelog, publishing, etc), or feel free to take over!


packages/machines/dialog/src/dialog.types.ts
```diff
@@ -69,6 +69,11 @@ type PublicContext = DirectionProperty & {
    * @default "dialog"
    */
   role: "dialog" | "alertdialog"
+  /**
+   * Whether to open or close the dialog on setup
+   * @default false
+   */
+  defaultIsOpen?: boolean
```
segunadebayo, line 76: Please use `isOpen` or `open` instead.

As much as we're trying to model things around Chakra UI. We try to keep Zag naming convention as close to DOM as possible.

packages/machines/dialog/src/dialog.machine.ts
```diff
@@ -26,15 +26,24 @@ export function machine(ctx: UserDefinedContext = {}) {
...
+                target: "closed",
+                guard: (ctx) => !ctx.defaultIsOpen,
+                actions: "setupDocument",
+              },
+              {
+                target: "open",
+                guard: (ctx) => !!ctx.defaultIsOpen,
+                actions: "setupDocument",
+              },
+            ],
```
segunadebayo, line 46: I think we can simplify this

```suggestion
            SETUP:  {
                target: ctx.open ? "open": "closed",
                actions: "setupDocument",
              },
```

segunadebayo (changes requested):

segunadebayo:
To generate a changeset for this PR, run `yarn changeset` and follow the prompts.

alexluong:
Thanks @segunadebayo for the feedbacks, made those changes + added example page for Solid and Vue + generated changeset

packages/machines/dialog/src/dialog.types.ts
```diff
@@ -69,6 +69,11 @@ type PublicContext = DirectionProperty & {
    * @default "dialog"
    */
   role: "dialog" | "alertdialog"
+  /**
+   * Whether to open or close the dialog on setup
+   * @default false
+   */
+  isOpen?: boolean
```
anubra266, line 76: I feel `defaultOpen` is better here since it's not controlled
```suggestion
  isOpen?: boolean
```

alexluong: I agree that it's probably better to have "default" somewhere in it as `isOpen` can be misinterpreted as a controlled variable.

segunadebayo: The notion of controller and uncontrolled is something we need to be cautious about as it’s a concept that’s generalized by React. 

I think we need to broaden our scope beyond React and see as if we’re building DOM primitives. Solid.js for example, doesn’t have the notion of controlled-uncontrolled since it compiles directly to the DOM (no VDOM)

Native APIs use naming like value, open, checked, etc to denote the initial state of the element. I recommend we stick with this approach.

alexluong: Makes sense, thanks for the explanation.

Anything else I can do to move the PR forward?

PR #120, opened 2022-06-04
feat(tags-input): api to set the value of a specific tag

anubra266:
api to set the value of a specific tag, based on the index of the tag

```js
api.setValueAtIndex(2, 'newValue')
```

### DIFF
--- a/packages/machines/tags-input/src/tags-input.machine.ts
+++ b/packages/machines/tags-input/src/tags-input.machine.ts
@@ -84,8 +84,8 @@ export function machine(ctx: UserDefinedContext = {}) {
         DELETE_TAG: {
           actions: ["deleteTag"],
         },
-        EDIT_TAG: {
-          actions: ["editTag"],
+        SET_VALUE_AT_INDEX: {
+          actions: ["setValueAtIndex"],
         },
         CLEAR_ALL: {
           actions: ["clearTags", "focusInput"],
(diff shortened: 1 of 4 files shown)


packages/machines/tags-input/src/tags-input.machine.ts
```diff
@@ -392,6 +395,12 @@ export function machine(ctx: UserDefinedContext = {}) {
           ctx.log.prev = ctx.log.current
           ctx.log.current = { type: "update", value: ctx.editedTagValue! }
         },
+        setValueAtIndex(ctx, evt) {
+          ctx.value[evt.index] = evt.value ?? ""
+          // log
+          ctx.log.prev = ctx.log.current
+          ctx.log.current = { type: "update", value: evt.value! }
+        },
```
segunadebayo, line 403: I think we should guard against `evt.value`, if it's not provided, we should warn and do nothing instead of pushing an empty string.

packages/machines/tags-input/src/tags-input.machine.ts
```diff
@@ -392,6 +396,16 @@ export function machine(ctx: UserDefinedContext = {}) {
           ctx.log.prev = ctx.log.current
           ctx.log.current = { type: "update", value: ctx.editedTagValue! }
         },
+        setValueAtIndex(ctx, evt) {
+          if (evt.value) {
+            ctx.value[evt.index] = evt.value ?? ""
```
segunadebayo, line 401:
```suggestion
            ctx.value[evt.index] = evt.value
```

packages/machines/tags-input/src/tags-input.machine.ts
```diff
@@ -392,6 +396,16 @@ export function machine(ctx: UserDefinedContext = {}) {
...
+            // log
+            ctx.log.prev = ctx.log.current
+            ctx.log.current = { type: "update", value: evt.value! }
```
segunadebayo, line 404:
```suggestion
            ctx.log.current = { type: "update", value: evt.value }
```

anubra266: chore: add Non-null assertion operator to `evt.value`

PR #123, opened 2022-06-07
fix: use guards in setup

anubra266:
Using guards for checks in Setup to avoid visualizer failing
CleanShot 2022-06-07 at 17 01 47@2x


packages/machines/popover/src/popover.machine.ts
```diff
@@ -41,10 +41,17 @@ export function machine(ctx: UserDefinedContext = {}) {
       states: {
         unknown: {
           on: {
-            SETUP: {
-              target: ctx.open ? "open" : "closed",
-              actions: ["setupDocument", "checkRenderedElements"],
```
segunadebayo, line 46: It seems you removed the `checkRenderedElements`?

PR #161, opened 2022-06-19
fix: prioritize combobox value over label

anubra266:
## 📝 Description

Use option values for component logic

Currently using option label


packages/machines/combobox/src/combobox.utils.ts
```diff
@@ -0,0 +1,10 @@
+import { dom } from "./combobox.dom"
+import type { MachineContext as Ctx } from "./combobox.types"
+
+export const utils = {
+  labelFromValue: (ctx: Ctx, value: string) => {
+    const el = dom.getMatchingOptionEl(ctx, value)
+    const data = dom.getOptionData(el)
+    return data.label
+  },
+}
```
segunadebayo, line 10: Please move this to the dom file and rename it to `getValueLabel`

.changeset/sweet-taxis-agree.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zag-js/combobox": patch
+---
+
+give value higher priority over label.
```
segunadebayo, line 5: Can you re-word this to explain the problem it solves?

packages/machines/combobox/src/combobox.connect.ts
```diff
@@ -34,7 +34,7 @@ export function connect<T extends PropTypes = ReactPropTypes>(state: State, send
     activeOption: state.context.activeOptionData,
     selectedValue: state.context.selectedValue,
     setValue(value: string) {
-      send({ type: "SET_VALUE", value })
+      send({ type: "SET_VALUE", value, isOptionValue: true })
```
segunadebayo, line 37: Any reason for the `isOptionValue: true` here?

segunadebayo (changes requested):
Thanks for this. Can you also run the tests to ensure it's all working?

packages/machines/combobox/src/combobox.connect.ts
anubra266: @segunadebayo the function that handles `setValue` also handles the input `onChange` so we need to differentiate when the event value is the option `value` from when it's the `label`, so when it's value, we can get the label first before changing the input value.

anubra266:
> Thanks for this. Can you also run the tests to ensure it's all working?
@segunadebayo Tests are working fine
CleanShot 2022-06-19 at 18 21 52@2x

PR #191, opened 2022-07-07
chore: Remove SETUP step

anubra266:
## Remove SETUP step and Improve SSR

Omit Setup step
 
- [x] Fix Visualizer output for machines with conditional targets, 
   - [x] Toggle
   - [x] Popover
   - [x] DIalog
   - [x] Checkbox


packages/utilities/dom/src/query.ts
```diff
@@ -1,3 +1,9 @@
+type Ctx<T extends string> = Partial<Record<T, () => Document | ShadowRoot | Node>>
```
segunadebayo, line 1: Let's move this closer to the withRootHelper function, and simplify.

```suggestion
type Ctx = { getRootNode?: () => Document | ShadowRoot | Node }
```

PR #203, opened 2022-07-14
build: update toolchain to pnpm + turbo + tsup

segunadebayo:
This PR removes a lot of the issues we had with `yarn` and its hoisting techniques.

- Switch from yarn to pnpm
- Update CI to use pnpm and composite github actions
- Simplify scripts directory
- Fix bugs in package dependencies (this most likely causes issues in the past releases)
- Remove our custom build script. Don't reinvent the wheel


segunadebayo (commented):
Self review: Good to go

PR #231, opened 2022-07-24
Add `setDisabled` method to API

### LINKED ISSUE #230: Slider cannot be dynamically disabled
# 🐛 Bug report

Dynamically changing the `disabled` option of the slider machine does not affect its enabled/disabled state. It seems like the only way to change it is to remount/rekey the slider with `disabled` set to the desired value. There are situations where this is less than ideal and could cause other bugs.

## 💥 Steps to reproduce

1. Set up slider per the docs
2. Set the `disabled` option of the slider machine to a boolean piece of state
3. Change the state from `false` to `true` or vice versa after the slider has mounted. Notice that the slider's enabled status is not affected

## 💻 Link to reproduction

CodeSandbox reproduction: https://codesandbox.io/s/billowing-forest-21bvc4?file=/src/App.tsx

## 🧐 Expected behavior

If `disabled` is dynamically changed to `true`, I expect the slider to be disabled and no longer respond to user events

## 🌍 System information

| Zag Version      |  0.1.13 react / 0.1.11 slider          |
| Browser          |   latest Chrome          |
| Operating System |     Mac OS Monterey       |

anubra266:
Closes #230

## 📝 Add `setDisabled` method

Adds the `setDisabled` method to API, to enable dynamic `disabled` state change in components.

### DIFF
--- a/.changeset/happy-horses-tie.md
+++ b/.changeset/happy-horses-tie.md
@@ -1,0 +1,8 @@
+---
+"@zag-js/checkbox": minor
+"@zag-js/rating": minor
+"@zag-js/slider": minor
+"@zag-js/toggle": minor
+---
+
+Add `setDisabled` method to API
--- a/.xstate/checkbox.js
+++ b/.xstate/checkbox.js
@@ -34,6 +34,9 @@ const fetchMachine = createMachine({
     SET_ACTIVE: {
       actions: "setActive"
     },
+    SET_DISABLED: {
+      actions: "setDisabled"
+    },
     SET_HOVERED: {
       actions: "setHovered"
     },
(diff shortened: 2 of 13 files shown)


segunadebayo:
Hi @anubra266,

For now, I think [this approach](https://zagjs.com/overview/programmatic-control#transient-updates) would suffice for a good amount of use cases.

In the future, I'd rather have us provide something like `api.setContext(...)` to allow the users to update some parts of the context. Let's wait till this becomes an urgent need.

anubra266:
> Hi @anubra266,
> 
> 
> 
> For now, I think [this approach](https://zagjs.com/overview/programmatic-control#transient-updates) would suffice for a good amount of use cases.
> 
> 
> 
> In the future, I'd rather have us provide something like `api.setContext(...)` to allow the users to update some parts of the context. Let's wait till this becomes an urgent need.

@segunadebayo how can we satisfy the issue in the linked issue

PR #249, opened 2022-08-10
Add RadioGroup Machine

anubra266:
Add radio component


packages/machines/radio/src/radio.types.ts
```diff
@@ -0,0 +1,118 @@
...
+   * If `true`, the radio input is marked as required,
+   */
+  required?: boolean
+  /**
+   * Defines the string that labels the checkbox element.
+   */
+  "aria-label"?: string
+  "aria-labelledby"?: string
+  "aria-invalid"?: boolean
+  "aria-describedby"?: string
```
segunadebayo, line 117: Do we really need all these props?

packages/machines/radio/src/radio.types.ts
```diff
@@ -0,0 +1,118 @@
...
+   */
+  disabled?: boolean
+  /**
+   * If `true`, the radio will be readonly
+   */
+  readonly?: boolean
+  /**
+   * If `true`, the radio is marked as invalid.
+   */
+  invalid?: boolean
```
segunadebayo, line 98: Let's update the naming and comments here. `active` > `activeId`

packages/machines/radio/src/radio.types.ts
anubra266: They're just accessibility attributes for each radio input, and I felt we should be consistent with the checkbox component too, because we have them there.

packages/machines/radio/src/radio.machine.ts
```diff
@@ -0,0 +1,101 @@
...
+          },
+        },
+      },
+    },
+
+    {
+      guards: {
+        isManual: (_, evt) => !!evt.manual,
+      },
+      activities: {
```
segunadebayo, line 58: This is unused. Let's remove it

segunadebayo (approved):
Looks good to me. Fix the comment above and we're good to go.

PR #306, opened 2022-09-06
tests: hover card machine

anubra266:
tests: hover card machine

### DIFF
--- a/e2e/hover-card.e2e.ts
+++ b/e2e/hover-card.e2e.ts
@@ -47,7 +47,7 @@ test("should be closed after blurring trigger", async ({ page }) => {
   await expect(page.locator(content)).not.toBeVisible()
 })
 
-test.only("should be closed after blurring trigger with keyboard", async ({ page }) => {
+test("should be closed after blurring trigger with keyboard", async ({ page }) => {
   await page.click("main")
   await page.keyboard.press("Tab")
   await page.waitForSelector(content)
--- a/e2e/hover-card.e2e.ts
+++ b/e2e/hover-card.e2e.ts
@@ -47,6 +47,18 @@ test("should be closed after blurring trigger", async ({ page }) => {
   await expect(page.locator(content)).not.toBeVisible()
 })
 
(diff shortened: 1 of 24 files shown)


e2e/hover-card.e2e.ts
```diff
@@ -0,0 +1,87 @@
...
+  await page.waitForSelector(content)
+  await expect(page.locator(content)).toBeVisible()
+})
+
+test("should be closed after blurring trigger", async ({ page }) => {
+  await page.focus(trigger)
+  await page.waitForSelector(content)
+  await expect(page.locator(content)).toBeVisible()
+
+  await page.locator(trigger).evaluate((e) => e.blur())
```
segunadebayo, line 45: Let's write this test as a user would use it.

Instead of calling blur, did you try to press the tab key multiple times?

examples/next-ts/pages/hover-card.tsx
```diff
@@ -25,25 +25,29 @@ export default function Page() {
   return (
     <>
       <main className="hover-card">
-        <a href="https://twitter.com/zag_js" target="_blank" {...api.triggerProps}>
-          Twitter
-        </a>
+        <div style={{ display: "flex", gap: "50px" }}>
+          <a href="https://twitter.com/zag_js" target="_blank" {...api.triggerProps}>
+            Twitter
```
segunadebayo, line 30: We need to edit the hover-card for the other frameworks?

e2e/hover-card.e2e.ts
anubra266: @segunadebayo I felt Tab is just one way to blur it. The user could just click outside also to blur

segunadebayo: Then we should have tests for both the keyboard blur and pointer outside blur as well.

anubra266: Done @segunadebayo

segunadebayo (approved):
Good job!

PR #318, opened 2022-09-15
feat(splitter): new splitter api

anubra266:
feat(splitter): new splitter api

### DIFF
--- a/packages/utilities/number/src/number.ts
+++ b/packages/utilities/number/src/number.ts
@@ -23,7 +23,8 @@ export const valueToPercent = (v: number | string, r: Num<"step" | "min" | "max"
 export const percentToValue = (v: number | string, r: Num<"min" | "max">) => r.min + (r.max - r.min) * valueOf(v)
 
 export function clamp(v: number | string, o: Num<"min" | "max">) {
-  return Math.min(Math.max(valueOf(v), o.min), o.max)
+  const min = Math.max(valueOf(v), o.min)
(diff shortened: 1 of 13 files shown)


shared/src/routes.ts
```diff
@@ -15,6 +14,7 @@ export const routesData: RouteData[] = [
   { label: "Nested Menu", path: "/nested-menu" },
   { label: "Menu With options", path: "/menu-options" },
   { label: "Context Menu", path: "/context-menu" },
+  { label: "Hover-card", path: "/hover-card" },
```
segunadebayo, line 17:
```suggestion
  { label: "Hover Card", path: "/hover-card" },
```

packages/utilities/number/src/number.ts
```diff
@@ -23,7 +23,8 @@ export const valueToPercent = (v: number | string, r: Num<"step" | "min" | "max"
 export const percentToValue = (v: number | string, r: Num<"min" | "max">) => r.min + (r.max - r.min) * valueOf(v)
 
 export function clamp(v: number | string, o: Num<"min" | "max">) {
-  return Math.min(Math.max(valueOf(v), o.min), o.max)
+  const min = Math.max(valueOf(v), o.min)
+  return typeof o.max === "undefined" ? min : Math.min(min, o.max)
```
segunadebayo, line 27: I think max being undefined should be handled from the caller, not within this function

segunadebayo:
Closing this for now till we figure out what to do.

PR #326, opened 2022-09-26
feat: add pagination

anubra266:
This PR introduces the new Pagination component machine to Zag.js. The goal is to use it within Chakra UI


segunadebayo (commented):
- [x] Rename `itemCount` to `count`
- [x] Rename `currentPage` to `page`
- [x] Rename pageSize to `perPage`

- [x] Please use .ts not .json for the pagination data
- [ ] onChange should take an object (We want all callbacks always to return an object)
- [x] Rename "dot" to "ellipsis"

packages/machines/pagination/src/pagination.utils.test.ts
```diff
@@ -0,0 +1,17 @@
...
+    expect(range.length).toEqual(5)
+  })
+
+  test("transform method", () => {
+    const items = [1, 2, "ellipsis", 3, 4]
+    const transformed = utils.transform(items)
+    expect(transformed[0].type).toEqual("page")
+    expect(transformed[2].type).toEqual("ellipsis")
+  })
+})
```
segunadebayo, line 17: For this test please use a snapshot instead

`expect(transformed).toMatchInlineSnapshot()`

PR #334, opened 2022-10-05
chore: rename messages to translations

anubra266:
rename messages to translations


packages/machines/tags-input/src/tags-input.types.ts
```diff
@@ -180,7 +180,7 @@ type PrivateContext = Context<{
   log: { current: Log | null; prev: Log | null }
   /**
    * @internal
-   * The live region to announce messages to the user
+   * The live region to announce translations to the user
```
segunadebayo, line 183:
```suggestion
   * The live region to announce changes to the user
```

PR #337, opened 2022-10-05
fix(editable): make component uncontrolled

anubra266:
Make editable component uncontrolled

### DIFF
--- a/packages/machines/editable/src/editable.connect.ts
+++ b/packages/machines/editable/src/editable.connect.ts
@@ -81,7 +81,6 @@ export function connect<T extends PropTypes>(state: State, send: Send, normalize
       "data-readonly": dataAttr(isReadonly),
       "aria-invalid": ariaAttr(isInvalid),
       "data-invalid": dataAttr(isInvalid),
-      value: state.context.value,
       size: autoResize ? 1 : undefined,
       onChange(event) {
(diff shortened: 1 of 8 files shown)


packages/machines/editable/src/editable.connect.ts
```diff
@@ -81,7 +81,6 @@ export function connect<T extends PropTypes>(state: State, send: Send, normalize
       "data-readonly": dataAttr(isReadonly),
       "aria-invalid": ariaAttr(isInvalid),
       "data-invalid": dataAttr(isInvalid),
-      value: state.context.value,
```
segunadebayo, line 84: I think we should change this to `defaultValue` instead of just removing it.

The reason is that setting a default value in the context will not reflect in the input if we remove it, making SSR output incorrect.

packages/machines/editable/src/editable.machine.ts
```diff
@@ -28,6 +29,10 @@ export function machine(ctx: UserDefinedContext) {
         },
       },
 
+      watch: {
+        value: ["invokeOnChange", "setInputValue"],
```
segunadebayo, line 33:
```suggestion
        value: ["invokeOnChange", "syncInputValue"],
```

Let's use a better name for that, so it's not confusing. There's setValue, and then there's setInputValue.

PR #367, opened 2022-10-18
fix(editable): dispatch custom events

anubra266:
Dispatch custom events in editable

### DIFF
--- a/.changeset/brown-steaks-smash.md
+++ b/.changeset/brown-steaks-smash.md
@@ -6,9 +6,8 @@ Dispatch custom events
 
 ```js
 useEffect(() => {
-  const cleanup = api.on("change", (evt) => {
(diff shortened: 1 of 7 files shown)


.changeset/brown-steaks-smash.md
```diff
@@ -0,0 +1,14 @@
...
+---
+
+Dispatch custom events
+
+```js
+useEffect(() => {
+  const cleanup = api.on("change", (evt) => {
+    console.log("Editable's new value is: ", evt.detail.value)
+  })
+  return () => cleanup()
```
segunadebayo, line 12:
```suggestion
  return api.on("change", (evt) => {
    console.log("Editable's new value is: ", evt.detail.value)
  })
```

PR #391, opened 2022-11-11
fix(checkbox): ssr issue

anubra266:
Fix SSR issue with Checkbox.
The reason this is happening is that we're checking defaultChecked after the machine starts, not in the initial state of the machine.

### DIFF
--- a/packages/machines/checkbox/src/checkbox.connect.ts
+++ b/packages/machines/checkbox/src/checkbox.connect.ts
@@ -29,7 +29,7 @@ export function connect<T extends PropTypes>(state: State, send: Send, normalize
 
   const trulyDisabled = isDisabled && !isFocusable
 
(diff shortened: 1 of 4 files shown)


packages/machines/checkbox/src/checkbox.connect.ts
```diff
@@ -29,7 +29,7 @@ export function connect<T extends PropTypes>(state: State, send: Send, normalize
 
   const trulyDisabled = isDisabled && !isFocusable
 
-  const stateView = state.hasTag("checked") ? "checked" : "unchecked"
+  const stateView = state.value || "unchecked"
```
segunadebayo, line 32: This is leaking the state directly into the view. In the event we add more state nodes, it'll show up in the `state.value` here. Let's be a bit more explicity

```suggestion
  const stateView = state.matches("checked") ? "checked" : "unchecked"
```

PR #400, opened 2022-11-21
chore: add `form` attribute to all form related elements

anubra266:
Add `form` attribute to all form related elements

### DIFF
--- a/packages/machines/pin-input/src/pin-input.types.ts
+++ b/packages/machines/pin-input/src/pin-input.types.ts
@@ -19,6 +19,10 @@ type PublicContext = DirectionProperty &
      */
     name?: string
     /**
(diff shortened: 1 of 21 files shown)


packages/machines/pin-input/src/pin-input.types.ts
```diff
@@ -19,6 +19,14 @@ type PublicContext = DirectionProperty &
      */
     name?: string
     /**
+<<<<<<< HEAD
```
segunadebayo, line 22: This seems like you pushed a merge conflict? Please resolve correctly

segunadebayo: Same for the other comments as well

PR #416, opened 2022-12-02
feat(anatomy): add ways to declare a machines anatomy

cschroeter:


packages/anatomy/src/anatomy.ts
```diff
@@ -0,0 +1,32 @@
...
+      _parts = [..._parts, ...parts]
+      return api
+    },
+
+    build() {
+      return [...new Set(_parts)].reduce<Record<string, { selector: string }>>(
+        (prev, part) =>
+          Object.assign(prev, {
+            [part]: {
+              selector: `[data-scope="${_component}"][data-part="${part}"]`,
```
segunadebayo, line 24: I would love to add these attributes as well to reduce the need for manually adding these to every component.

```suggestion
              selector: `[data-scope="${_component}"][data-part="${part}"]`,
              attrs: { "data-scope":  _component, "data-part": part }
```

packages/machines/accordion/src/accordion.anatomy.ts
```diff
@@ -0,0 +1,3 @@
+import { anatomy as createAnatomy } from "@zag-js/anatomy"
+
+export const anatomy = createAnatomy("accordion").parts("root", "item", "trigger", "content")
```
TimKolberger, line 3: **suggestion**: Using a unique name for each component prevents name clashes when exporting and helps auto import.

```suggestion
export const accordionAnatomy = createAnatomy("accordion").parts("root", "item", "trigger", "content")
```

packages/anatomy/src/anatomy.ts
```diff
@@ -0,0 +1,32 @@
+export const anatomy = (component: string) => {
```
TimKolberger, line 1: **thought (non-blocking)**: instead of renaming later on, we could rename the function

```suggestion
export const createAnatomy = (component: string) => {
```

packages/machines/accordion/src/accordion.anatomy.ts
TimKolberger: Ah nice, thanks!

packages/anatomy/src/create-anatomy.ts
```diff
@@ -0,0 +1,41 @@
+import { isEmpty } from "@zag-js/utils"
```
segunadebayo, line 1: Can we inline this function? Since it's the only function imported?

examples/solid-ts/vite.config.ts
```diff
@@ -6,14 +6,5 @@ export default defineConfig({
...
     target: "esnext",
-    polyfillDynamicImport: false,
-  },
-  resolve: {
-    alias: [
-      {
-        find: /^@zag-js\/(.*)$/,
-        replacement: path.resolve("./node_modules/@zag-js/$1/src"),
-      },
-    ],
```
segunadebayo, line 17: We added these so we won't have to build the project before running it. I'm curious to know what your intentions are here.

cschroeter: Restored

packages/anatomy/src/create-anatomy.ts
cschroeter: Sure.

PR #433, opened 2022-12-08
fix: focus on tab

### LINKED ISSUE #428: [Select] Pressing Tab on open select moves focus to wrong element
# 🐛 Bug report

Pressing "Tab" when a select is open results in the wrong element being focused.

## 💥 Steps to reproduce

1. Click on select trigger to open the select menu
2. Press "Tab"
3. The button next to the select trigger does not receive focus

## 💻 Link to reproduction

CodeSandbox reproduction: https://codesandbox.io/s/sharp-scooby-tlq38i?file=/src/App.vue

## 🧐 Expected behavior

The button next to the select trigger receives focus

## 🧭 Possible Solution

Move focus to the trigger element and let browsers handle default tab behavior.

## 🌍 System information

| Zag Version      |  0.1.2  |
| Browser          |  Chrome 107.0.5304.121  |
| Operating System |  MacOS  |

visualjerk:
Closes #428 

## 📝 Description

Fixes focus handling when tabbing out of an open select.

Focus did not move to the next / previous tabbable element.

## 🚀 New behavior

Focus moves to the next / previous tabbable element.

No

## Considered Alternative

Keeping `event.preventDefault` and recreating the browser behavior when tabbing out of an element. This turns out to be pretty hard, as it would involve focusing browser control elements when reaching the first/last tabbable element inside the dom.


segunadebayo:
Hi @visualjerk. I'm really happy to see that you took the plunge to fix the issue. Thank you!

The solution works on the first try, and I love it.

The only issue I think we need to test is this:
- Add a focusable element within the select (say an input)
- When the select menu opens, move focus to the input instead
- Press the tab key
- Expect that the focus behavior works the same as expected

If we can confirm that this works, I'm happy to merge this in.

visualjerk:
Hi @segunadebayo! In order to proceed with this, I need some clarification about these steps:

> * Add a focusable element within the select (say an input)
> * When the select menu opens, move focus to the input instead

Should focus be moved to the focusable element automatically or by clicking on the it?

segunadebayo:
Hi @visualjerk,

Sorry, I've been taking some time to think more about this. I believe the Select component should trap focus and block scroll. This will eliminate the need to resolve this logic.

visualjerk:
@segunadebayo thanks for the update. In that case, is there anything left to do for this MR?

segunadebayo:
Closing this based on my replied here:
https://github.com/chakra-ui/zag/pull/433#issuecomment-1359110047

PR #492, opened 2023-01-31
feat/vue scoped machines

codebender828:
- fix: add jsx preserve to tssconfig
- chore: cleanup toast example

Closes # 

## Problem statement (Current behaviour)
The problem this attempts to solve is that if the `useMachine` hook is called outside the component's lifecycle, then the state machine is never initialized. This means that the developer, would need to create a global variable, create a local state machine API and then assign it to the global variable via side-effects before it can be used in a global context.

Furthermore, if the component's lifecycle is destroyed, then the state machine is also destroyed along with it. This would leave the global variable referencing an already disposed variable.

## 📝 Description

> Add a brief description

This PR introduces a proof of concept for component machine initialization outside framework component lifecycles but can still execute inside any component's runtime or in vanilla JS.

I encountered a scenario in Vue where I might need to extract or initialize the machine API outside the component context for the toast machine.

### Proposed solution:

Add an option to the `useMachine` hook to define the runtime scope of the component machine.
```ts
// vue/useMachine.ts
export type MachineRuntimeScope = "global" | "component"
```

So I made a few modifications to the `useMachine` and `useService` hooks to allow the consumer to define where they want the machine execution context to begin:

`global` means, it should initialize and be disposed in the window’s runtime scope. `component` means it should initialize and be disposed of within the component’s runtime scope.

That way, we can initialize the machine state outside the component context and can still use it without depending on the `onMounted` or `useEffect` hooks to initialize the component state machines.

For example, in with the toast, we may want to initialize and invoke the `toastApi.create` method from outside the Vue app. or before the Vue app is initialized. Here is the effect of the `MachineRuntimeScope` option of `useMachine` would produce some of the effect below:

https://user-images.githubusercontent.com/21237954/215740033-29ddbcfc-2db5-427b-863b-657921928629.mp4

## 🚀 New behavior

- Existing functionality remains as-is.
- Component machines can now execute in the global runtime context if the `global` option is passed.

## 📝 Additional Information
This PR implements this functionality inside of the Vue.js framework hook. I think this this is less than ideal but I'm open to feedback here.

I think it would be better handled inside a separate package that could perhaps be shared across multiple frameworks, or even just vanilla JS.

An example usage in Vanilla JS would be something to this effect:
In vanilla JS I imagine it would look something like this:

```ts
// in main.ts script

// import vanilla JS defineMachine
import { defineMachine } from "@zag-js/vanilla"

// import the component machine
import * as toast from "@zag-js/toast"

function makeToaster() {
  const [state, send] = toast.group.machine({ id: "toast.group" }), {
    context: {
      pauseOnInteraction: true,
    }
  }

  const toaster = toast.group.connect(state, send)
  return toaster
}



const toaster = makeToaster()

// Event listeners
const button = document.querySelector(".some-button")
button.addEventListener("click", () => {
  toaster.create({
    title: "Here's some toast!"
  })
}}

```


segunadebayo:
Closing in favor of using the `useActor` hook as explained in Slack.

Thanks for sharing this detailed proposal.

Learn more about the actor model here:
https://xstate.js.org/docs/packages/xstate-vue/#api

PR #522, opened 2023-02-10
Draft: combobox tags

visualjerk:
## 📝 Description

This is a draft for composing combobox and tags input as discussed here: #435 

Currently there is only a Vue example and it might contain some bugs or strange behavior.

No intend to merge this one soon, but would appreciate some feedback about the composition. In particular the following topic comes to mind:

### Preventing Tags Input Idle State

As the combobox's popover will most probably be rendered in a teleport, clicking on it will trigger the tags input "blur" event. This results in the tags input entering idle state, which prevents further interaction (e.g. selecting tags with arrow keys).

The current solution is to add an additional check to the tags input `trackInteractOutside` activity:

```ts
// handle composition when used as combobox
const combobox = document.getElementById(`${ctx.id}:combobox:listbox`)
return contains(dom.getRootEl(ctx), target) || contains(combobox, target)
```

This feels quite hacky, as we bind the tags input machine to internals of the combobox.

A better solution might be to add an additional context prop to the tags input, that extends the `trackInteractOutside` scope.

Do you have any preferences or ideas on that topic?

## 🚀 Next Steps

- [ ] Add E2E tests for the composition
- [ ] Cleanup example
- [ ] Add example pages for react and solidjs

No


packages/machines/tags-input/src/tags-input.machine.ts
```diff
@@ -268,7 +268,9 @@ export function machine(userContext: UserDefinedContext) {
         trackInteractOutside(ctx, _evt, { send }) {
           return trackInteractOutside(dom.getInputEl(ctx), {
             exclude(target) {
-              return contains(dom.getRootEl(ctx), target)
+              // handle composition when used as combobox
+              const combobox = document.getElementById(`${ctx.id}:combobox:listbox`)
+              return contains(dom.getRootEl(ctx), target) || contains(combobox, target)
             },
             onInteractOutside() {
               send({ type: "BLUR", src: "interact-outside" })
```
segunadebayo, line 276: Good attempt @visualjerk. Love to see your approach.

How about we allow the users to control the interact outside logic from the context. 

```jsx
onInteractOutside(e) {
     if(!ctx.isInteractionOutside?.(e)) return
     send({ type: "BLUR", src: "interact-outside" })
}
```

visualjerk: Looks like an even better approach, as it gives users more flexibility. I have created a PR for this #533

segunadebayo:
What's the status of this PR @visualjerk?

visualjerk:
Regarding Zag functionality, all blockers have been resolved, so the only thing holding me back is a shortage of time. Have been quite busy with work and a holiday trip lately. I hope to find time to complete this next month.

segunadebayo:
Got it @visualjerk. I'll close this for now. 

Please create a new PR when you have it ready.

Thanks for working on this.

PR #533, opened 2023-02-17
feat(tags): expose interact outside in context

visualjerk:
## 📝 Description

This adds an `onInteractOutside ` hook to tags input context, to enable composition with other components (e.g. `combobox`).

Related PR: #522

## 🚀 New behavior

By adding an `onInteractOutside ` to the tags input context, users can hook into the logic, which determines if an event should be treated as an outside interaction.

Example:

```ts
import { useId } from "react"
import * as tagsInput from "@zag-js/tags-input"

tagsInput.machine({
  id: useId(),
  onInteractOutside(event) {
    if (comboboxContentEl.contains(event.target)) {
      event.preventDefault()
    }
  },
})
```

No

## 📝 Additional Information

As it is somewhat tedious to cover this with an e2e test, I opted for unit testing the state machine. This might be a good starting point for implementing a broader test suite, which provides faster feedback.

Please let me know, if you consider this to be a direction worth going. Otherwise I am also fine with adding e2e tests instead.


packages/machines/tags-input/src/tags-input.machine.ts
```diff
@@ -270,7 +270,8 @@ export function machine(userContext: UserDefinedContext) {
             exclude(target) {
               return contains(dom.getRootEl(ctx), target)
             },
-            onInteractOutside() {
+            onInteractOutside(event) {
+              if (ctx.isInteractionOutside && !ctx.isInteractionOutside(event)) return
```
segunadebayo, line 274: Looking at the current implementation, I found that calling `event.preventDefault()` on the event passed here will also prevent the function from being called.

So, we could consider doing:

```js
onInteractOutside(event) {
    ctx.onInteractOutside?.(event)
    if (event.defaultPrevented) return
    send({ type: "BLUR", src: "interact-outside" })
},
```

segunadebayo:
I love that you brought in unit testing here since it's a more realistic way of testing it.

With the approach, I'd recommend that we render the full example of the component in the `setupTest` function to mirror actually usage scenario. You get what I mean?

packages/machines/tags-input/src/tags-input.machine.ts
visualjerk: Looks like a more sensible approach. Just to doublecheck, is this event a custom event provided by `trackInteractOutside` or the actual event which triggered it listens to? In the later case, using `event.preventDefault()` in the context's `onInteractOutside` would also prevent the browsers default behavior, which might be unexpected.

visualjerk:
Regarding the unit test setup: Do you mean a full example with `tags-input.connect` and one of the use packages (e.g. `@zag-js/react` alongside `@testing-library/react`? If that is the case, should we create a separate package for these tests or rather add the dependencies to the workspace root package?

packages/machines/tags-input/src/tags-input.machine.ts
segunadebayo: Yes, it's a custom event provided by interaction outside, so preventing default is fine.

segunadebayo:
Good question @visualjerk. Let me think about the test setup a bit more.

It's a tricky one because we've got multiple frameworks

visualjerk:
My initial idea was to focus on unit testing just the state machine, but I mixed it up with a bit too much dom integration 😄  If we want more of a unit test approach, mocking out everything related to the dom might be the way to go.

If we want to treat these tests more like faster e2e tests, we might consider creating a dedicated folder for them. We can then have a dom setup for each framework and a single test suite, which runs the same tests for every framework. So it would be pretty close to the cypress e2e tests.

visualjerk:
Added a few changes to look into an e2e / integration test suite with vitest. With this setup it should be easy to run the test suite for all frameworks. With this approach the tests are less reliant on internal state, while still providing faster feedback as the real e2e tests.

visualjerk:
Actually it turned out to be tricky having both react and vue test setups in one package. The only working solution I found is having a vite config for each framework.

As some types for JSX are conflicting there is however still a type error in `tags-input.setup.vue.tsx`.

segunadebayo:
Yeah, I faced this as well.

To move this task forward, we should start by adding the `onInteractOutside` to the context for most of the related components. Let's get that merged first, then we can figure out the tests later

visualjerk:
Sounds like a good strategy. I created a new branch for `onInteractOutside`. Closing this in favor of #553

PR #536, opened 2023-02-22
refactor: split up dom utils

segunadebayo:
- Refactor codebase for consistent dependency structure
- Organizes the packages and deprecates the `dom-utils` package
- Creates more single purpose packages


segunadebayo (commented):
Looks good. Now, let's test to see that it works!

PR #550, opened 2023-03-01
feat: improve support for svelte

anubra266:
Trying out svelte through actions based on this: https://github.com/chakra-ui/zag/discussions/262#discussioncomment-5155884

Issues after first try

- [x] I tried accordion, and there seems to be a render issue where the items all flash on initial render, very possible such will happen for other components.
- [x] The sample uses sveltekit, there seems to be a problem with initial render, where zag is trying to access the browser before it's available, so it throws a 500 - **Most likely SSR**


segunadebayo:
Geneally speaking, it's a good idea to have Svelte in there.

It's SSR I'm mostly concerned with here, I understand that i can be achieved with actions but they don't work in SSR.

I think Zak might have some solutions for Svelte SSR already. Please reach out to him

jescowuester:
We could use the actions only for event handlers and spread the remaining props like normally. 
Would be a bit less pretty, but should get rid of the ssr issues.

segunadebayo:
That would be amazing @jescowuester. Looking forward to a POC for this.

Shyrro:
Hey @anubra266 !

I wanted to push something to your PR but wasn't able to. So instead i created this repo : https://github.com/Shyrro/test-zag-svelte

I made the spread work with the accordion machine. I'm still having reactivity issues but those are mainly coming from what i did and not related to SSR or anything. I'm missing reactivity somewhere and didn't dig further. 
As you can see when launching the repo, the props are spread correctly, the events also have no issue being used correctly. 

Your code was good, and i think your code is working(i just changed some stuff for the useMachine and normalizeProps), but the main issue i had in the zag repo, is that for some reason, importing packages from the workspace doesn't seem to work.  
In the push i wanted to make, i downgraded the version for the accordion machine, that way i explicitly download the npm version instead of the workspace one and i didn't have anymore issues. 

Maybe @segunadebayo can help resolving these? 

I will fork the repo as soon as i can to also better emphasize this

jescowuester:
What's the current status for this? Anything I can do to help out at this point?

Shyrro:
Hey @jescowuester ! Thanks for your help, your PR for splitting attributes and handlers was very helpful !

We still need to handle some types so that when using the Svelte normalizer we have the right typings, then add some docs and we should be good to go for Svelte !

jescowuester:
There was still one open bug that I had on my previous PR: 

Popovers aren't positioned correctly. This happens because there are two sources for the `styles` we apply:
- The "style" attribute in the props we spread on the component 
- The styles that are being set via JS by zag itself when the popover position is calculated
```

Unlike in other frameworks, the "style" attribute that we spread onto an element overwrites _all_ styles for that element in svelte (https://github.com/sveltejs/svelte/issues/455). 


the issue with the dialog jumping around is that we set styles on the element directly


But the attributes we pass to a component also include a style prop. This props overwrites all the styles on the element (https://github.com/sveltejs/svelte/issues/455), so the position styles get lost.

I assume the styles are currently not passed via props directly for performance reasons, is that right?

**Comment by Shyrro** [comment]
Hi @jescowuester 

Thanks for this. Didn't notice the issue. 

Let me know if i'm wrong, but considering the Svelte issue you mentionned, wouldn't be better to pass a string to the style attribute like this `style="color: red;transform: val"` ? Spreading won't work and obviously resets the styles, but writing a string like this should behave like the native HTML style prop if i'm not mistaken ?

I am not a fan of the setAttributes onMount/onUpdate, as it is not optimized and we would have to handle all the styles ourselves. 

I'll try to work something out and see if i can come with an acceptable solution.

**Comment by Shyrro** [comment]
Hey @jescowuester and @endigma 

I just tried the styles with the current version ( converting the styles to a string ) and it seems to work perfectly without overriding the styles. 

I implemented the menu and it works fine, look at this : 

[Menu gif positionner](https://s2.gifyu.com/images/ezgif-1-6951921601.gif)

We can see that the Menu positions itself and that the styles are not overriden by the style. 

Let me know if you notice another issue! Thanks for your help!

**Comment by anubra266** [comment]
@Shyrro I guess we can just have a method in `normalizeProps` that converts styles to string.

**Comment by Shyrro** [comment]
@anubra266 Already there 😉

File: packages/frameworks/svelte/src/zag.ts [hunk 1]
```diff
@@ -0,0 +1,13 @@
+export function zag(node: Element, handlers: Record<string, () => void>) {
```
segunadebayo, line 1: This action feels more like `events` not `zag` since there's nothing zag-related about it.

> <div use:events={api.rootProps.handlers} />

segunadebayo:
@anubra266 @Shyrro. Let's get this branch merged ASAP.

Don't want us to have a long-running branch

anubra266:
@segunadebayo sure. Other changes will be in subsequent pull requests

PR #554, opened 2023-03-01
chore: add integration tests

visualjerk:
## 📝 Description

Adds first integration tests for tags-input.

-

## 🚀 New behavior

-

No

## 📝 Additional Information

As discussed in #533, this is an effort to add integration tests with vitest. The idea is to slowly build additional test suites that:

- provide more immediate feedback than playwright tests by running faster
- provide direct access to a machines context per test
- simplify testing edge cases

Currently the test suite only includes a setup for react. I think this might be enough, if we see these tests as an addition to real e2e tests with playwright. The main focus would be on testing shared machines.

### DIFF
--- a/integration/package.json
+++ b/integration/package.json
@@ -1,0 +1,23 @@
+{
+  "private": true,
+  "scripts": {
+    "test": "vitest"
+  },
+  "dependencies": {
+    "@testing-library/dom": "9.0.0",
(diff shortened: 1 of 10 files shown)


segunadebayo:
Closing this for now to reduce the maintenance burden. 

I think we can do without it for now.

PR #578, opened 2023-03-21
fix(types): fix handlers and attributes types

Shyrro:
## 📝 Description

> Add TS support for Svelte normalizer

Current behavior returns errors on TS with the "attributes" and "handlers". This modifications aims to solve that.

## 🚀 New behavior

> When using the svelte normalizer, we expect the return to be an object with "attributes" and "handlers" instead.

No

### DIFF
--- a/packages/frameworks/svelte/src/get-attributes-and-handlers.ts
+++ b/packages/frameworks/svelte/src/get-attributes-and-handlers.ts
@@ -1,6 +1,8 @@
-export function getAttributesAndHandlers(props: Record<string, any>) {
-  const handlers = {} as Record<string, any>
-  const attributes = {}
+import type { Dict } from "@zag-js/core/src/types"
+
+export function getAttributesAndHandlers(props: Dict) {
+  const handlers: Dict = {}
+  const attributes: Dict = {}
 
   Object.keys(props).forEach((key) => {
     // this is probably not a safe way to distiguish handlers
--- a/packages/frameworks/svelte/src/get-attributes-and-handlers.ts
+++ b/packages/frameworks/svelte/src/get-attributes-and-handlers.ts
@@ -1,5 +1,5 @@
 export function getAttributesAndHandlers(props: Record<string, any>) {
-  const handlers = {}
+  const handlers = {} as Record<string, any>
   const attributes = {}
 
   Object.keys(props).forEach((key) => {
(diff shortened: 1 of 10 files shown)


packages/types/src/prop-types.ts
```diff
@@ -44,15 +48,17 @@ type DataAttr = {
   [key in `data-${string}`]?: string | number | Booleanish
 }
 
-export type PropTypes = Record<"button" | "label" | "input" | "output" | "element" | "select" | "style", Dict>
+export type PropTypes<T = any> = Record<
+  "button" | "label" | "input" | "output" | "element" | "select" | "style",
+  T extends SplitArgs ? SplitArgs : Dict
+>
```
segunadebayo, line 54: How about using the `T` type param and setting it to `Dict` by default?

```suggestion
export type PropTypes<T = Dict> = Record<
  "button" | "label" | "input" | "output" | "element" | "select" | "style",
  T
>
```

packages/types/src/prop-types.ts
```diff
@@ -1,6 +1,10 @@
 import type { JSX } from "./jsx"
 
 type Dict<T = any> = Record<string, T>
+export type SplitArgs<T = any> = {
+  attributes: Record<string, T>
+  handlers: Record<string, EventListener>
+}
```
segunadebayo, line 7: Let's move this to the svelte package if possible

packages/frameworks/svelte/src/get-attributes-and-handlers.ts
```diff
@@ -1,5 +1,5 @@
 export function getAttributesAndHandlers(props: Record<string, any>) {
-  const handlers = {}
+  const handlers = {} as Record<string, any>
```
segunadebayo, line 2:
```suggestion
  const handlers: Dict = {}
```

Shyrro:
Hey @segunadebayo ! 

Thanks for your feedback 🙏 

I pushed a commit containing the changes asked. Let me know if i need to change something else. 😊 

PS:  with Abraham we noticed that `attributes` and `handlers` coming from the Svelte normalizer, are lightly typed compared to other frameworks. For example in react `getRootProps` will have the attributes returned by the normalize function(eg: "tabindex"), whereas in Svelte the attributes type will be `Record<string, any>`. 
I am still looking for a way to improve this. Let me know if you want me to find a solution for this before moving on to the rest with Svelte. Thanks again 🙏

PR #589, opened 2023-04-01
feat: add svelte kit app

Shyrro:
## 📝 Description

Managed to make a SvelteKit app work within the pnpm workspace and resolve the deps + HMR correctly.

Adding a SvelteKit app to the repo didn't work because of deps
## 🚀 New behavior

Deps now work in SvelteKit app in our pnpm workspace thanks to `ssr.noExternal`.

## 📝 Additional Information

We now have a SvelteKit App to test the machines against. So we can develop the machines in confidence for SSR mode


anubra266:
Looks great. So we can't use `workspace:*` instead of `link`

Shyrro:
@anubra266 We can! We just need to include those packages in "ssr noExternal", otherwise Vite doesn't know how to resolve them in  ssr

anubra266:
@Shyrro I guess we just need to get rid of the current sandbox, and ensure the shared styling works here, then we can merge into the svelte branch.

PR #590, opened 2023-04-02
feat: svelte improve and add menu

Shyrro:
## 📝 Description

This PR contains the following : 

- Fix for style attribute ( now included in the use directive rather than the native spread )
- Fix for handlers not being updated correctly ( we used to attach them once the connect is init, but they need to be updated for each change of the api )
- Renamed `attributes` to `attrs` since lines are long enough
- Renamed `zag` method to `attach` to better fit its purpose
- Add Menu component example

Currently, sometimes, there are noticeable bugs with the style attribute when the styles are set from elsewhere. 
The handlers were only set once instead of each time the api changes.

## 🚀 New behavior

- Instead of returning `{attributes, handlers}`, we now return `{attributes, rest}`. `rest` contains everything that isn't in the attributes(handlers and styles for now)
- Handlers are now refreshed on each api update
- Styles are handled via "setProperty" instead of spread

No

## 📝 Additional Information

- Tested the following under the SvelteKit sandbox
 - Mouse, Keyboard events and selection for the Menu
 - Mouse, Keyboard events and selection for the Accordion
 - Mouse, Keyboard events and selection for the Select


anubra266:
@jescowuester we would love to your input on the changes we're making here.

examples/svelte-kit/src/routes/accordion/+page.svelte
```diff
@@ -16,25 +16,25 @@
 </script>
 
 <main class="accordion">
-  <div use:events={api.rootProps.handlers} {...api.rootProps.attributes}>
```
segunadebayo, line 19: I'm not sure I dig this change.

- If we have only event handlers in `rest`, then it makes sense to keep the `handlers` naming so it's more precise.
- Same logic applies to `use:events`, because it's mostly event handlers that is going to use that action, I'd recommend keeping that name

Except I'm missing some more context behind the name change. I believe this syntax feels more intuitive:

```jsx
<div use:events={api.rootProps.handlers} {...api.rootProps.attrs}>
```

Shyrro: @segunadebayo Thanks for reviewing this !

We have both `handlers` and `style` in `rest`, not just handlers. 

We need to handle the style attribute in the directive so that it doesn't override the styles coming from floating UI. If we don't, the popovers placements are a bit off.

I couldn't come up with a better naming, let me know what you think !

PR #624, opened 2023-05-03
Update package.json

MrHBS:
Since noEmit is set to true in tsconfig.json, do you still need to pass this flag?


segunadebayo:
Hi @MrHBS,

It's okay to keep it in event the root tsconfig `noEmit` is changed at some later time.

PR #629, opened 2023-05-09
refactor: switch machine, add switch test

junghyeonsu:
## 📝 Description

As with [commit 415509](https://github.com/chakra-ui/zag/commit/41550949e16b9f4879b89ed7e3658271912ebfaf), I fixed machine and transition in switch.

Same with before

## 🚀 New behavior

- I added a test for switch, which is no different than checkbox, so I configured it the same.
- Removed `line-height: 0` that was being held by shared css on switch. I removed it because the test was failing because it didn't find the label when I tested it.
- I've added tests to test it, and I've tested it myself for each framework in the examples, so let me know if you see anything that doesn't work.

No

### DIFF
--- a/.xstate/switch.js
+++ b/.xstate/switch.js
@@ -1,42 +1,45 @@
-"use strict"
+"use strict";
 
-var _xstate = require("xstate")
-const { actions, createMachine, assign } = _xstate
-const { choose } = actions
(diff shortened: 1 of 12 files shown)


segunadebayo:
Good job @junghyeonsu.

Thanks for working on this.

PR #630, opened 2023-05-09
fix(transient-transitions): support watchable context

Shyrro:
Add support for reactive context as second option of `useMachine`

### DIFF
--- a/packages/frameworks/svelte/src/use-machine.ts
+++ b/packages/frameworks/svelte/src/use-machine.ts
@@ -1,12 +1,12 @@
 import { onMount } from "svelte/internal"
 import type { MachineSrc, StateMachine as S } from "@zag-js/core"
-import { writable } from "svelte/store"
+import { Readable, derived, writable } from "svelte/store"
 
 export function useService<
   TContext extends Record<string, any>,
   TState extends S.StateSchema,
   TEvent extends S.EventObject = S.AnyEventObject,
->(machine: MachineSrc<TContext, TState, TEvent>, options?: any) {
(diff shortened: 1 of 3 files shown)


packages/frameworks/svelte/src/use-machine.ts
```diff
@@ -1,12 +1,12 @@
...
 
 export function useService<
   TContext extends Record<string, any>,
   TState extends S.StateSchema,
   TEvent extends S.EventObject = S.AnyEventObject,
->(machine: MachineSrc<TContext, TState, TEvent>, options?: any) {
+>(machine: MachineSrc<TContext, TState, TEvent>, options?: S.HookOptions<TContext, TState, TEvent>) {
   const { actions, state: hydratedState, context } = options ?? {}
   const _machine = typeof machine === "function" ? machine() : machine
   const service = context ? _machine.withContext(context) : _machine
```
segunadebayo, line 12: Could we design the transient context here to use a writable store in Svelte (and the store could be an object)

Shyrro: You mean instead of having multiple stores inside the context, that the context itself is a store? 

We could, it would put more effort on the consumer side as a store doesn't automatically changes when one of its dependencies does. 
The consumer would have to track the dependencies himself and update the context. 

It's really just a design choice, i thought it'd be easier to use if we track the deps for the consumer.

I don't know if what i said was clear, let me know if i need to detail the approach further or draft a sandbox with a demo!

anubra266:
We'll be going with the #635 approach instead

PR #635, opened 2023-05-13
feat(svelte): add writable context

Shyrro:
This is in retrospect to what we talked about in the comment of #630 

It handles the transient context as a writable

### DIFF
--- a/examples/svelte-kit/src/stores/controls/ui.svelte
+++ b/examples/svelte-kit/src/stores/controls/ui.svelte
@@ -1,8 +1,6 @@
 <script lang="ts">
-  import { useControls } from "."
-
   export let controls: Record<string, any>
-  export let context: ReturnType<typeof useControls>[0]
+  export let context: any
 </script>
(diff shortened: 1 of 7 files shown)


examples/svelte-kit/src/stores/controls/ui.svelte
```diff
@@ -14,22 +14,23 @@
...
+          id={label}
           data-testid={key}
           type="text"
           {placeholder}
-          value={$ctx[key]}
+          value={$context[key]}
           on:keydown={(e) => {
             if (e.key === "Enter") {
-              ctx.update((s) => ({ ...s, [key]: e.currentTarget.value }))
+              context.update((s) => ({ ...s, [key]: e.currentTarget.value }))
```
Shyrro, line 33:
```suggestion
              $context = { ...$context, [key]: e.currentTarget.value }
```

doesn't this work? I believe it should since we're in a Svelte file

PR #666, opened 2023-05-29
fix: require not working in `svelte-kit` example

elcharitas:
## 📝 Description

I came across this issue a lot while trying to work on a svelte preprocessor for zag. As a result I figured it's better to fix this first.

An error shows up anytime one tries to navigate any of the components. It comes up because `country-list` uses require internally.

## 🚀 New behavior

The error no longer popsup

## 📝 Additional Information

A single dependency was added to make this possible.


anubra266:
Thanks @elcharitas

PR #676, opened 2023-06-02
Vanilla events normalization

Shyrro:
This PR is a POC that aims to resolve the change event not being triggered with vanilla JS. 

This triggers the change event when it is triggered programmatically rather than by the user. 

Currently, this is tailored for the checkbox, but we'd have to handle this generally for all the input elements that might trigger changes, such as `textarea` and the different input types such as `radio`, `text` etc.

Programmatic events not triggered a change event in svelte

## 🚀 New behavior

Programmatic "change" event, triggered when the property "checked" of the input changed.

PS: This also contains a small fix for the playground as SvelteKit seems to have trouble with CJS packages such as the form serializer and the select data. 
I removed the serializer in svelte, and used the json data directly instead of calling the cjs function for the Select.

### DIFF
--- a/examples/svelte-kit/src/routes/checkbox/+page.svelte
+++ b/examples/svelte-kit/src/routes/checkbox/+page.svelte
@@ -30,7 +30,6 @@
         id="check-id"
         disabled={api.isChecked}
         on:click={() => {
-          console.log("in")
           api.setChecked(true)
         }}
       >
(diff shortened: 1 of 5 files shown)


anubra266:
@Shyrro someone already fixed the cjs errors with a vite plugin. You might need to pull from main to see the changes.

segunadebayo:
Closing this as discussed

PR #768, opened 2023-07-27
Fix removeDelay docs inconsistency for toasts

dasmikko:
## 📝 Description

The toast documentation states that there is added 500ms of delay when a toast is removed, but the codes doesn't reflect that.

the removeDelay is always 0ms

## 🚀 New behavior

removeDelay is 500ms by default

No


segunadebayo:
Good catch! The code is correct, it's the docs that needed an update.

I've pushed an update for this.

PR #826, opened 2023-08-28
refactor: redesign select and combobox

segunadebayo:
## 📝 Description

This PR redesigns the select and combobox APIs to achieve the following:

- Support for passing value as strings
- Support for controlled open/closed states
- Improved APIs for controlling the select component


segunadebayo (commented):
Self Review: Looks good!

PR #828, opened 2023-08-29
fix: dispatch change event in Checkbox and Switch

srflp:
## 📝 Description

While using Checkbox (and the same goes for Switch) with an external form library (namely [Felte](https://felte.dev/), with SolidJS) I noticed that the Felte form is not getting the updated checkbox value after I change it programmatically using `api().setChecked(value)`.

The reason for that is that Felte is listening for `change` events, yet zag dispatches only a `click` event after any change of Checkboxes value.

A brief table displaying current behavior and requested changes:
|  For the input type: | checkbox/radio/file/hidden | text and the rest |
|--------|--------|--------|
| Zag currently dispatches an event of type | `click` | `input` |
| Felte listens for the event of type | `change` | `input` | 
| Zag after this change dispatches events of type | `click`, `change` | `input` | 

For input type="checkbox", Zag currently dispatches an event of type `click`.

## 🚀 New behavior
For input type="checkbox", Zag dispatches an event of type `click` and `change`.

No

## 📝 Additional Information
Felte event handling code:
https://github.com/pablo-abc/felte/blob/0a9941032dcbef7285f114be04470cbef4b4bc5c/packages/core/src/create-form-action.ts#L208

### DIFF
--- a/.changeset/wild-students-rush.md
+++ b/.changeset/wild-students-rush.md
@@ -1,0 +1,5 @@
+---
+"@zag-js/form-utils": patch
+---
+
+Dispatch change event in Checkbox and Switch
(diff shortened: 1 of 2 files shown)


cschroeter:
Lgtm

segunadebayo:
Thanks for working on this @srflp.

I've gone ahead to apply this fix to even more components and refactored it a bit.

Pushed this to the main branch already.

PR #844, opened 2023-09-08
fix(select): issue with solid

cschroeter:


segunadebayo:
I managed to fix this in the main branch. Thanks for catching that.

Issue #885, opened 2023-09-10
Dialog closes when a toast is visible and clicked, despite closeOnOutsideClick set to false


damzobridge:
### Description

When a `toast` is triggered within the `Dialog`, any touch or click with the toast causes the `Dialog` to close, even though the `closeOnOutsideClick` prop is set to false on the Dialog component.

When I trigger a toast within the dialog, I expected the dialog to remain open until I explicitly close it, but it closes unexpectedly upon interacting with the toast.

### Link to Reproduction

https://codesandbox.io/p/sandbox/elastic-oskar-5vnq2k

### Steps to reproduce

1. Go to the provided CodeSandbox link.
2. Click on the `Open Dialog` button to open the dialog.
3. Within the dialog, click on the `Add Toast` button to trigger a toast.
4. Click on the toast.
5. Observe that the dialog closes unexpectedly.

### Ark UI Version

0.14.0

### Framework

- [X] React
- [ ] Solid
- [ ] Vue

segunadebayo:
I just pushed a fix for this. We'll release an update shortly.

If the issue persists after upgrading, I'll re-open it.

PR #889, opened 2023-09-21
fix(dialog): do not close Dialog on outside focus

srflp:
improvement for #885

## 📝 Description

Do not close Dialog on outside focus.

At first I thought I would solve this like so:
```ts
onFocusOutside(event) {
  if (!ctx.closeOnOutsideClick) event.preventDefault()
},
```

But then I thought that there is no case of user wanting to close the Dialog by focusing anything else. So I left it like this:
```ts
onFocusOutside(event) {
  event.preventDefault()
},
```

This is an incomplete fix, since I think that clicking on Alerts shouldn't close Dialogs at all, no matter whether the closeOnOutsideClick flag is set to true or false. With closeOnOutsideClick equal to true, Dialogs should get closed only when one clicks on the overlay, not other top-level components.

Dialog **is** closed when anything outside of Dialog is getting focused. 
When closeOnOutsideClick is set to false, the Dialog gets closed when Alert's button gets focused by clicking it. 

## 🚀 New behavior

Dialog **is not** closed when anything outside of Dialog is getting focused. 
When closeOnOutsideClick is set to false, the Dialog does not get closed when Alert's button gets focused by clicking it. 

No

### DIFF
--- a/packages/machines/dialog/src/dialog.machine.ts
+++ b/packages/machines/dialog/src/dialog.machine.ts
@@ -82,6 +82,9 @@ export function machine(userContext: UserDefinedContext) {
               }
               ctx.onEsc?.()
             },
+            onFocusOutside(event) {
+              event.preventDefault()
+            },
             onPointerDownOutside(event) {
               if (!ctx.closeOnOutsideClick) {
                 event.preventDefault()


segunadebayo:
Good call @srflp!

In an attempt to fix the same issue, I made some changes to the dialog code that affects your PR.

I think the underlying fix is to ensure clicking on a toast doesn't close the dialog, but I also. made it possible for the user to control this behavior, to be more explicit.

PR #923, opened 2023-10-05
docs(combobox): add combobox properties to docs and examples

srflp:
## 📝 Description

Add misssing Combobox properties on website and in examples.

CleanShot 2023-10-05 at 20 21 36

No


segunadebayo:
Thanks for working on this @srflp

I don't think we should overwhelm the users with these many options. We only need to document them, which we don't do a good job of right now.

I've gone ahead to cherry-pick the necessary changes we need to make.

PR #936, opened 2023-10-12
feat(toast): remove removeDelay

cschroeter:


segunadebayo:
Tried to get it to the finish line, but it's taking longer than it's worth.

We'll stick with the remove delay for now. I'll definitely revisit this in the future.

PR #950, opened 2023-10-19
fix(combobox): handle loop property correctly

srflp:
## 📝 Description

Sooo.. support for loop property in Combobox got deleted at some point. Combobox ignored the loop property and always looped. I reimplemented the loop property and added some tests which test the Combobox with loop property equal to false.

Combobox is looping even when loop=false.

## 🚀 New behavior

Combobox is looping only when loop=true.

No

## 📝 Additional Information

Additionally I remove a duplicated item from a combobox list example and fix some typos.


.xstate/combobox.js
```diff
@@ -129,7 +129,7 @@ const fetchMachine = createMachine({
           actions: ["highlightFirstSelectedItem", "invokeOnOpen"]
         }, {
           target: "interacting",
-          actions: ["highlightNextItem", "invokeOnOpen"]
```
srflp, line 132: I'm changing this to FirstItem in ARROW_DOWN since in ARROW_UP there is LastItem.

PR #985, opened 2023-11-03
fix(select): use optional chaining operator for safe access to ctx's highlightedItem

acd02:
## 📝 Description

On my previous PR (#954), I've introduced a bug, sorry about that 🙇
I didn't check if `highlightedItem` from the context was defined before trying to access its value.

When defining the guards for `CONTENT.ARROW_UP` (or `CONTENT.ARROW_DOWN`), I put `isFirstItemHighlighted` (or `isLastItemHighlighted`) at the end, thinking they would be called after the `hasHighlightedItem` guard, which checks if `highlightedItem` is indeed defined.
I wrongly assumed that if one guard fails, there would be an early return and the subsequent guards wouldn't be called, but that's not the case it seems.

```tsx
{
  guard: and("hasHighlightedItem", "isLoopEnabled", "isFirstItemHighlighted"),
  actions: ["highlightLastItem"],
},
```
This PR fixes that by using the optional chaining operator to safely access the `highlightedItem` value

No

### DIFF
--- a/e2e/select.e2e.ts
+++ b/e2e/select.e2e.ts
@@ -254,3 +254,17 @@ test.describe("select / focused / select option", () => {
     await expectToBeChecked(page.locator(getOption("KE")))
   })
 })
+
+test.describe("select / loop / keyboard", () => {
+  test("should loop through the options when loop is enabled", async ({ page }) => {
+    await controls(page).bool("loop")
+    await page.focus(trigger)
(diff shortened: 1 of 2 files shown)


segunadebayo:
Hi @acd02. Thanks for the follow-up PR.

I've cherry-picked the changes and applied them to the main branch.

PR #986, opened 2023-11-04
feat(machines): collapsible

Omikorin:
## 📝 Description

Add a Collapsible machine for managing height/width calculations


packages/machines/collapsible/src/collapsible.connect.ts
```diff
@@ -44,7 +44,7 @@ export function connect<T extends PropTypes>(state: State, send: Send, normalize
       "aria-expanded": isOpen,
       // hidden: !isOpen,
       style: {
-        "--collapsible-height": `${state.context.height}px`,
+        height: isOpen ? `${state.context.height}px` : "0px",
```
segunadebayo, line 47: - We should expose this as a css var.
- We should also support height/width collapsing as well

PR #1132, opened 2024-01-03
fix: pin input paste behaviour

renrizzolo:
## 📝 Description

When pasting in the pin input after it has been filled, the copied text will sometimes fill a single input, where it should be limited to a single character. This is mostly a visual bug, as the state value is still 'correct' (the first char is used).

Also fixes some inconsistency with pasting a value after being filled. It should now paste the copied text from the current focused input onward, without appending the focused input's value.

https://github.com/chakra-ui/zag/assets/13904763/a2e23a4b-bc25-4a20-a44b-0cc124bf72ce

https://github.com/chakra-ui/zag/assets/13904763/a00246f1-b7e5-47a9-b09d-10f97ee4ac30

## 🚀 New behavior

https://github.com/chakra-ui/zag/assets/13904763/608f8bfd-4a98-44a3-908d-d65f1b869aa9

No

### DIFF
--- a/packages/machines/pin-input/src/pin-input.connect.ts
+++ b/packages/machines/pin-input/src/pin-input.connect.ts
@@ -105,6 +105,12 @@ export function connect<T extends PropTypes>(state: State, send: Send, normalize
               send({ type: "VALUE.INVALID", value })
               event.preventDefault()
             }
+
+            // select the text so paste always replaces the
+            // current input's value regardless of cursor position
+            if (value.length > 2) {
+              event.currentTarget.setSelectionRange(0, 1, "forward")
+            }
           } catch {
             // noop
           }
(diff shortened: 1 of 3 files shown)


packages/machines/pin-input/src/pin-input.connect.ts
```diff
@@ -105,6 +105,12 @@ export function connect<T extends PropTypes>(state: State, send: Send, normalize
               send({ type: "VALUE.INVALID", value })
               event.preventDefault()
             }
+
+            // select the text so paste always replaces the
+            // current input's value regardless of cursor position
+            if (value.length > 2) {
+              event.currentTarget.setSelectionRange(0, 1, "forward")
+            }
```
renrizzolo, line 113: this isn't totally exhaustive, the buggy behavior can still happen when pasting a single character. I don't think there's a way to detect a paste in `onBeforeInput`.

PR #1159, opened 2024-01-17
feat: svelte adapter part 1

izznatsir:
## 📝 Description

Svelte adapter with Svelte 5.

Currently, there is no Svelte adapter due to Svelte 4 and below limitation around spreading event handlers.

## 🚀 New behavior

Svelte 5 remove that limitation.

No

## 📝 Additional Information

Still work in progress, todos:

- [x] Generate type declaration on build.
- [x] Write examples.
   - [x] Accordion
   - [x] Avatar
   - [x] Carousel
   - [x] Checkbox
   - [x] Color Picker
- [ ] Write docs.


segunadebayo:
Love the progress you're making with this @izznatsir 👌🏼✨

Thanks for working on it

gbouteiller:
@izznatsir Hello, I was working on the same thing on my own. I've got nearly the same result as you so maybe we should join forces and talk about the different technical choices (even if svelte 5 is not here yet 🙃).

izznatsir:
> @izznatsir Hello, I was working on the same thing on my own. I've got nearly the same result as you so maybe we should join forces and talk about the different technical choices (even if svelte 5 is not here yet 🙃).

Maybe we should merge this PR as is. Then, if you have different technical perspective you can submit a PR and discuss there. I'm afraid the size of PR will be to big if including all examples and docs.

izznatsir:
But, it should be excluded from publish script until it's truly ready.

segunadebayo:
@izznatsir, Happy to get this merged ASAP and we'll set the svelte package to "private" for now.

Let me know when it's good to go.

izznatsir:
Upstream bugs:

- Unstate in derived removing a nested prop value [svelte #10315](https://github.com/sveltejs/svelte/issues/10315)
- Stale reactive prop [svelte #10316](https://github.com/sveltejs/svelte/issues/10316)

izznatsir:
@segunadebayo It's ready for review. Thanks.

anubra266:
@izznatsir There's conflicts, please update your lockfile

examples/svelte-ts/src/routes/Carousel.svelte
```diff
@@ -0,0 +1,33 @@
...
+  import Toolbar from "$lib/components/Toolbar.svelte"
+  import { useControls } from "$lib/use-controls.svelte"
+  import * as carousel from "@zag-js/carousel"
+  import { carouselControls, carouselData } from "@zag-js/shared"
+  import { normalizeProps, useMachine } from "@zag-js/svelte"
+
+  const controls = useControls(carouselControls)
+
+  const machine = useMachine(carousel.machine({ id: "1", index: 0, spacing: "20px", slidesPerView: 2 }), {
+    // context: controls.context,
```
anubra266, line 11: Is preventing disturbing this from working?

izznatsir: It works for the default context, but when I change slidesPerView to 2 in controls UI, it skip second slide when I click next. It also occurs at next-ts example so I commented it out.

izznatsir: It works for the default context, but when I change slidesPerView to 2 in controls UI, it skip second slide when I click next. It also occurs at next-ts example so I commented it out.

gbouteiller: About the behavior of context in the second argument, a point was made in another issue on Solid that it seemed to only work on React.  I also asked for the use of this second argument but didn't get any response yet. I think a svelte 5 way to deal with reactive context  is :
```ts
let {context} = $props<Props>();
const machine = useMachine(accordion.machine({id: '1', value: ['Aircrafts'], ...context}));
let api = $derived(accordion.connect(machine.state, machine.send, normalizeProps));
$effect(() => machine.service.setContext(context)); // <= HERE
```

izznatsir: The custom context is working, please see use-machine.svelte.ts file. It use the same logic as you have described above.

Here, I think it is an issue in the carousel logic when being controlled as it also happens in next-ts example.

gbouteiller: Yes, I saw your code, the context is working indeed because it is turned into a state by your custom function called useControls so the effect in use-service.svelte can work. My approach is different as I get context from $props (from a $derived in parent component), that's why I manage the reactivity here. Anyway, potahto-potato 😊. As you say, this one seems to be related to the machine itself.

PR #1161, opened 2024-01-19
fix(dialog): open was not controlled by context

### LINKED ISSUE #1027: [React][Dialog] Controlled Dialog is not fully controlled
### Description

Hi!

When using the `Dialog` component in controlled mode, there is an issue with the close behavior. When in controlled mode, I expect to fully control "how" the dialog will be closed but that's not the case currently as the dialog will always be closed internally when clicking on the `CloseTrigger` component, using the Esc key, or clicking on the backdrop.

As a workaround, I could listen for the `onOpenChange` property and update the parent state, but it does not make any sense as  the open state of the dialog should be "controlled" by the parent directly. Also, it causes an extra render because we're trying to sync both the state of the parent and the internal dialog state.

When looking at other component libraries (MUI, HeadlessUI, etc.), a dialog in controlled mode won't be closed unless the parent sets the proper `open` value.

https://codesandbox.io/s/ark-ui-dialog-controlled-2l28fg?file=/src/Dialog.tsx

### Steps to Reproduce

1. Go to the reproduction link
2. Open the dialog
3. Close the dialog by clicking on the close button
4. Check the console logs, you should see a message indicating that the `onOpenChange` hander has been called with the `false` value meaning the internal dialog state has changed
5. Try to open the dialog again
6. The dialog won't open because the parent state is not in sync.

**Expected behavior**
The dialog should not close in this case when I click on the close button because I don't set the `open` props to `false`.

### Ark UI Version

1.0.1

### Framework

- [X] React
- [ ] Solid
- [ ] Vue

### Browser

Google Chrome 119

### Additional Information

It's possible that we have the same issue with other components that have a controlled mode and overlay.

juliomuhlbauer:
Closes #1027

## 📝 Description

The dialog connect was not using the context open value, so it was working the being controlled.

Not updating machine on open value change.

## 🚀 New behavior

Tracking the open value on the dialog's connect. But not working with back and forth navigation on next :(

No

## 📝 Additional Information
Before:

https://github.com/chakra-ui/zag/assets/53458125/6ec8d2a6-a9f4-42fb-9625-6039b20e6598

https://github.com/chakra-ui/zag/assets/53458125/a5fa83e9-7e14-4329-91fe-c272361c59eb

After:

https://github.com/chakra-ui/zag/assets/53458125/74c96c81-d13b-4c61-9cc0-4515e93534af

https://github.com/chakra-ui/zag/assets/53458125/15808df5-b506-431f-89a2-549c984f36d3


packages/machines/dialog/src/dialog.connect.ts
```diff
@@ -5,7 +5,7 @@ import type { MachineApi, Send, State } from "./dialog.types"
 
 export function connect<T extends PropTypes>(state: State, send: Send, normalize: NormalizeProps<T>): MachineApi<T> {
   const ariaLabel = state.context["aria-label"]
-  const isOpen = state.matches("open")
+  const isOpen = state.matches("open") || !!state.context.open
```
segunadebayo, line 8: This isn't sufficient to fix the issue for the following reasons.

- It only modified the DOM `hidden` value, but the machine might still be in the open state, even though it shows as visually closed.

- Because the machine state might be out of sync, the focus trap and scroll lock features might still be activated. Could you try adding more content on the page to confirm this?

The ideal solution will involve creating new "open:controlled" and "close:controlled" states for the machine

juliomuhlbauer: I am having some trouble understanding how the machine works.

image

There is already the open and closed state in the dialog machine. It would be nice having some docs on how the codebase works for newbies.


Also I added some height in the example and it works fine.

juliomuhlbauer:
Seems to be fixed

PR #1183, opened 2024-01-25
feat: add data-disabled, data-invalid, aria-invalid, data-readonly to indicatorProps for Select

iNetJoJo:
Closes # 

## 📝 Description

> this will enable styling of indicator when building custom select component with chakra-ui and applying visual styles trough theme via: _disabled, _readOnly, _invalid props

> data and aria attributes are currently not visible on indicator atom

## 🚀 New behavior

> now those props are applied as compared to triggerProps

No

## 📝 Additional Information

indicatorProps are not shown in documentation, i saw that ark-ui uses it in its select implementation.

### DIFF
--- a/packages/machines/select/src/select.connect.ts
+++ b/packages/machines/select/src/select.connect.ts
@@ -205,6 +205,10 @@ export function connect<T extends PropTypes, V extends CollectionItem = Collecti
       dir: state.context.dir,
       "aria-hidden": true,
       "data-state": isOpen ? "open" : "closed",
+      "data-disabled": dataAttr(isDisabled),
+      "data-invalid": dataAttr(isInvalid),
+      "aria-invalid": isInvalid,
+      "data-readonly": dataAttr(isReadOnly),
     }),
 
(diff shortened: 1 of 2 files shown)


packages/machines/select/src/select.connect.ts
```diff
@@ -205,6 +205,10 @@ export function connect<T extends PropTypes, V extends CollectionItem = Collecti
       dir: state.context.dir,
       "aria-hidden": true,
       "data-state": isOpen ? "open" : "closed",
+      "data-disabled": dataAttr(isDisabled),
+      "data-invalid": dataAttr(isInvalid),
+      "aria-invalid": isInvalid,
```
segunadebayo, line 210: We don't need `aria-invalid` here

```suggestion
      "aria-invalid": isInvalid,
```

iNetJoJo: Not a deal breaker, but what if we want to change color of indicator i case a select field is invalid?

Or is this some semantic conflict?

PR #1209, opened 2024-02-01
fix: DatePicker shows correct number of weeks when `startOfWeek` set

### LINKED ISSUE #1208: [DatePicker] Number of weeks wrong when `startOfWeek` set
# 🐛 Bug report

The DatePicker calculates the number of weeks to show based on the locale, but it also accepts a `startOfWeek` prop which may differ from the locale's week start day. 

The Ark UI demo at https://ark-ui.com/docs/components/date-picker has `startOfWeek` set to 1, but the weeks are calculated based on weeks starting with Sunday for my locale.

<table>
<tr>
 <td>

image
 <td>

image
</table>

## 💥 Steps to reproduce

1. Use the DatePicker component
2. Set `startOfWeek` to a number which doesn't match the locale.
3. Check each month, some may have extra weeks and some may have weeks cut off
4. For me, setting `startOfWeek` to 1 causes Oct 2023 and Dec 2023 to display incorrectly.

## 💻 Link to reproduction

CodeSandbox reproduction: https://codesandbox.io/p/sandbox/peaceful-leaf-rnvhsn?file=%2Fsrc%2FApp.tsx%3A13%2C40

## 🧐 Expected behavior

The calendar should show the correct dates when `startOfWeek` is set.

## 🧭 Possible Solution

Either calculate weeks based on the `startOfWeek` prop, or use the `locale` to determine the correct `startOfWeek`.

## 🌍 System information

| Zag Version      |    0.34.0     |
| Browser          |      Firefox      |
| Operating System |      Windows      |

zhengkyl:
Closes #1208

## 📝 Description

This adds a function that correctly calculates the number of weeks when the firstDayOfWeek does not match the locale.

Currently, the number of weeks only factors in the locale and ignores the firstDayOfWeek parameter. 

## 🚀 New behavior

Now, it uses the firstDayOfWeek if set and defaults to the locale otherwise.

No

## 📝 Additional Information

Another possible solution is to not allow passing in a custom `firstDayOfWeek` and only compute it from the locale. I'm not sure if there is a good reason to allow arbitrary week starts.


packages/utilities/date-utils/src/get-month-days.ts
```diff
@@ -16,8 +16,18 @@ export function getDaysInWeek(weekIndex: number, from: DateValue, locale: string
   return dates
 }
 
+export function getCustomWeeksInMonth(from: DateValue, locale: string, firstDayOfWeek?: number) {
+  if (firstDayOfWeek == null) {
+    return getWeeksInMonth(from, locale)
+  }
+  const paddingDays = (getDayOfWeek(startOfMonth(from), "en") - firstDayOfWeek + 7) % 7
```
segunadebayo, line 23: I believe this should use the locale instead of the hardcoded 'en'? 

```suggestion
  const paddingDays = (getDayOfWeek(startOfMonth(from), locale) - firstDayOfWeek + 7) % 7
```

zhengkyl: `getDayOfWeek` returns a number depending on locale. So `0` always matches the locale's week start. https://github.com/adobe/react-spectrum/blob/main/packages/%40internationalized/date/src/queries.ts#L72

The `firstDayOfWeek` and other code expects 0 to always be Sunday, so I chose "en" but it can be arbitrary. I couldn't figure out an simpler alternative since all the internationalized functions are relative to a locale. I can add a comment explaining.

PR #1256, opened 2024-02-17
fix: combobox empty input should not be replaced by selected value when typing

izznatsir:
## 📝 Description

On Vue and Svelte, empty input will be replaced by selected value when typing.

Empty value is replaced by selected value in `combobox.connect` function call. Because Vue and Svelte call `combobox.connect` each time state is updated in derive and do not have `defaultValue` concept like in react, empty input value keep being replaced while typing.

## 🚀 New behavior

Do not replace empty value with selected value while typing.

No

## 📝 Additional Information
Svelte combobox example is not yet available in this branch, but you can run the e2e-vue tests to check this fix.


segunadebayo:
Good catch @izznatsir

I'm merging this with #1323 to keep it in a single PR.

PR #1302, opened 2024-02-28
fix: make sure onFileReject is called when no files are accepted

Pagebakers:
Closes # 

## 📝 Description

> Add a brief description

## 🚀 New behavior


packages/machines/file-upload/src/file-upload.machine.ts
```diff
@@ -126,6 +126,8 @@ export function machine(userContext: UserDefinedContext) {
           if (acceptedFiles.length) {
             const files = ref([acceptedFiles[0]])
             set.files(ctx, files, rejectedFiles)
```
Munaf-Divan, line 128: Is there a specific reason to set `rejectedFiles` when file is accepted? @Pagebakers

Pagebakers: There can be accepted and rejected files if you select multiple files, so this should be correct.

PR #1303, opened 2024-02-28
docs(showcase): add `multiple` prop toggle to combobox docsite playground

TylerAPfledderer:
Addresses #1279


segunadebayo:
I see little value in adding this to the website. Multiple is useful when composing it with another component (like the TagsInput) or a chip group.

Since we don't have such a composition in place, let's close this for now.

PR #1304, opened 2024-02-29
feat: add isReadOnly prop and handle read-only state in switch component

nonoakij:
## 📝 Description

The documentation has a ReadOnly option, but it is not really available.

screenshot
https://zagjs.com/components/react/switch

## 🚀 New behavior

When set to ReadOnly, the value will not change. 
However, unlike disabled, the input element is not "disabled".

### DIFF
--- a/.changeset/brown-birds-explain.md
+++ b/.changeset/brown-birds-explain.md
@@ -1,0 +1,5 @@
+---
+"@zag-js/switch": minor
+---
+
+Updated readOnly props to work as expected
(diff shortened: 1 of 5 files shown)


segunadebayo:
Hi @nonoakij, I think it should be other way around instead. We should remove that section from the docs since we no longer support read only on checkboxes.

This is not supported by the native DOM hence the reason we removed it.

nonoakij:
@segunadebayo Thanks for the reply.

> We should remove that section from the docs since we no longer support read only on checkboxes.

I will close this pull request and create a new pull request to remove it from the docs

PR #1385, opened 2024-04-01
fix(tabbable): correct some returned types

ocavue:
~~Closes #~~

## 📝 Description

This PR simply fixes some TypeScript types from the package `@zag-js/tabbable`, by explicitly adding returned types. 

The current `.d.ts` file in [`@zag-js/tabbable@0.41.0`](https://unpkg.com/browse/@zag-js/tabbable@0.41.0/dist/index.d.ts) declares that some API returns `HTMLElement`. 

```ts
/**
 * Returns the first focusable element within the element
 */
declare function getFirstTabbable(...): HTMLElement;
```

## 🚀 New behavior

But they should return `HTMLElement | null`. 

```ts
/**
 * Returns the first focusable element within the element
 */
declare function getFirstTabbable(...): HTMLElement | null;
```
 
No

## 📝 Additional Information

N/A

### DIFF
--- a/.changeset/spicy-carpets-tell.md
+++ b/.changeset/spicy-carpets-tell.md
@@ -1,0 +1,5 @@
+---
+"@zag-js/tabbable": patch
+---
+
+Correct some returned types.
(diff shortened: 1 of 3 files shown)


segunadebayo:
Thanks for this PR @ocavue.

I love your work on Aria UI; it looks exciting. I wonder if you would be interested in contributing to Zag.js so we can support web components? Using preact-signals is a clever approach.

PR #1394, opened 2024-04-04
fix(select): keyboard interaction is not acceptable when focus is returned to trigger element

erm1116:
Closes # 

## 📝 Description
In the Select component, if the focus is returned to the trigger button (e.g. shift+tab) when the listbox is open, the listbox cannot be operated with the keyboard. 

https://github.com/chakra-ui/zag/assets/38889285/5921d5ee-4ade-4861-89a3-1daee01e07cb

## 🚀 New behavior

https://github.com/chakra-ui/zag/assets/38889285/f5734534-cdc8-4768-8775-8accc22f7df1

### DIFF
--- a/packages/machines/select/src/select.machine.ts
+++ b/packages/machines/select/src/select.machine.ts
@@ -9,6 +9,7 @@ import { addOrRemove, compact, isEqual } from "@zag-js/utils"
 import { collection } from "./select.collection"
 import { dom } from "./select.dom"
 import type { CollectionItem, MachineContext, MachineState, UserDefinedContext } from "./select.types"
+import { addDomEvent } from "@zag-js/dom-event"
 
 const { and, not, or } = guards
 
@@ -234,7 +235,13 @@ export function machine<T extends CollectionItem>(userContext: UserDefinedContex
           tags: ["open"],
           entry: ["focusContentEl"],
           exit: ["scrollContentToTop"],
-          activities: ["trackDismissableElement", "computePlacement", "scrollToHighlightedItem", "proxyTabFocus"],
+          activities: [
+            "trackDismissableElement",
+            "trackTriggerFocus",
+            "computePlacement",
+            "scrollToHighlightedItem",
+            "proxyTabFocus",
+          ],
           on: {
             "CONTROLLED.CLOSE": [
               {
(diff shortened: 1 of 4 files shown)


erm1116:
I made mistake that machine state when focus returned to trigger is not appropriate. I'll fix it

packages/machines/select/src/select.machine.ts
```diff
@@ -420,6 +427,27 @@ export function machine<T extends CollectionItem>(userContext: UserDefinedContex
             },
           })
         },
+        trackTriggerFocus(ctx, _evt, { send, getState }) {
```
erm1116, line 430: At first I tried to add `TRIGGER.FOCUS` event transition in the `open` state, but this causes state transition problem.
Specifically, event order when clicking trigger button is as belows:
1. `TRIGGER.CLICK`(sent by onPointerdown): idle -> open
2. `TRIGGER.FOCUS`(sent by onFocus): open -> focused

So eventually state is `focused` and it is not intended. (`open` state is expected)


Secondly, I thought that fixing the exclude  of `trackDismissableElement` is one of the approach.
But trigger button is focusable, `ctx.restoreFocus` will be false and transitions to `idle` state.
https://github.com/chakra-ui/zag/blob/a47da98c51801da0313bb6e4078dfa8203872793/packages/machines/select/src/select.machine.ts#L409-L420
https://github.com/chakra-ui/zag/blob/a47da98c51801da0313bb6e4078dfa8203872793/packages/machines/select/src/select.machine.ts#L296-L315


Finally, I added the tracker that listening trigger button focus event and send `CLOSE` event manually if it is focused in `open` state.

*note: addDomEvent is within raf funciton because listening focus event immediately, also causes the state transition problem.

segunadebayo:
I appreciate this PR @erm1116.

I'm starting to think we should not allow tabbing out of the select content when it's open. This matches the native select behavior and works similarly to Radix UI, for example.

segunadebayo:
Closing in favor of [30528d0](https://github.com/chakra-ui/zag/commit/30528d04563aa428d044867ce40c3ccab7cf8b7b)

erm1116:
@segunadebayo 
> I'm starting to think we should not allow tabbing out of the select content when it's open. This matches the native select behavior and works similarly to Radix UI, for example.

I partially agree with you.
I think Select component is like single and not editable Combobox component (like [Select-Only example in APG](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/)). 
In that case, the focus should remain on the trigger button when open state, and [Combobox should be in the Tab sequence](https://arc.net/l/quote/kfwlwduh) so I think not allowing tabbing out is unintended behavior.

How you think this?

—

FYI: When [I referred to this](https://discord.com/channels/964648323304808488/964648323304808496/1207913628402065490), maintainer consider as `Select component is organized as a Listbox widget with a trigger`, so maybe we don't need to follow Combobox's guide.

PR #1483, opened 2024-05-06
fix(collection): ensure collection are considered different when disabled change

maastrich:
## 📝 Description

Ensure that collection with attributes that changes (e.g. disabled) should be considered as changed.

When the items of a collection are updated with only a state change, the collection is considered as equal (as the values are the same).

## 🚀 New behavior

When the items of a collection are updated with only a state change, the collection should be considered as changed, as of now, the disabled attribute changed IS considered as a change for the collection.

No

### DIFF
--- a/packages/utilities/collection/src/collection.ts
+++ b/packages/utilities/collection/src/collection.ts
@@ -69,7 +69,7 @@ export class Collection<T extends CollectionItem = CollectionItem> {
       const label = this.itemToString(item)
       const disabled = this.itemToDisabled(item)
 
-      hashSet.add(value)
+      hashSet.add(JSON.stringify({ value, disabled }))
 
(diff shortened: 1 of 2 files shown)


packages/utilities/collection/src/collection.ts
```diff
@@ -69,7 +69,7 @@ export class Collection<T extends CollectionItem = CollectionItem> {
       const label = this.itemToString(item)
       const disabled = this.itemToDisabled(item)
 
-      hashSet.add(value)
+      hashSet.add(JSON.stringify({ value, disabled }))
```
maastrich, line 72: not sure of this specific form but I thought that `hashSet.add(value + disabled)` could create conflicts if a value ends with `true` or `false`

PR #1641, opened 2024-07-02
refactor: change top-level process call in proxy.ts

baptisteArno:
Related to https://github.com/chakra-ui/ark/discussions/2609

Avoid calling `process` at top level to avoid crash in web environment (where process is not defined)

### DIFF
--- a/.changeset/metal-falcons-boil.md
+++ b/.changeset/metal-falcons-boil.md
@@ -1,0 +1,5 @@
+---
+"@zag-js/store": minor
+---
+
+Improve reliability in web runtime by wrapping the process.env check in a function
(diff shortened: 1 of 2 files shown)


Jonas-C:
Isn't this be more of a bundling-related problem for zag-js specifically? Shouldn't the dev-related code be treeshaken before zag-js is bundled and published?

baptisteArno:
Thanks!

I mean, it might be also great to be consistent in the code base? Here, `isDev` is a function as well: https://github.com/chakra-ui/zag/blob/992bc3df40d4fc08fc467544b8b5effebe313e6c/packages/utilities/core/src/guard.ts#L1

segunadebayo:
I think it's fine to add this PR. Thanks for working on this @baptisteArno

@Jonas-C Yes, that's the intent. We don't want to add that env replacement on our end since we lose the warning/error features.

PR #1696, opened 2024-07-26
fix: improve is* checkers compatibility

ocavue:
~~Closes #~~ 

## 📝 Description

> Add a brief description

This PR includes two improvements for `isHTMLElement()` functions etc.
 
1. Ensure they work on Node.js, where global `Node` is not available. 
2. Check whether the past argument is a valid object. This ensures that passing a `null` or `undefined` won't throw errors. The typescript signature for past `ev` is `any` so it's better to support `null` and `undefined` etc.

```
$ echo "require('@zag-js/dom-query').isHTMLElement({});" | node

var isHTMLElement = (v) => typeof v === "object" && v?.nodeType === Node.ELEMENT_NODE && typeof v?.nodeName === "string";
                                                                    ^
ReferenceError: Node is not defined
```

```
$ echo "require('@zag-js/dom-query').isNode(null);" | node

var isNode = (el) => el.nodeType !== void 0;
                        ^
TypeError: Cannot read properties of null (reading 'nodeType')
```

## 🚀 New behavior

No error for both two examples above.

No

## 📝 Additional Information

N/A

### DIFF
--- a/packages/utilities/dom-query/src/is.ts
+++ b/packages/utilities/dom-query/src/is.ts
@@ -1,18 +1,12 @@
-const ELEMENT_NODE: typeof Node.ELEMENT_NODE = 1
-const DOCUMENT_NODE: typeof Node.DOCUMENT_NODE = 9
-const DOCUMENT_FRAGMENT_NODE: typeof Node.DOCUMENT_FRAGMENT_NODE = 11
+export const isHTMLElement = (v: any): v is HTMLElement =>
+  typeof v === "object" && v?.nodeType === Node.ELEMENT_NODE && typeof v?.nodeName === "string"
 
-const isObject = (v: unknown): v is Record<string, unknown> => typeof v === "object" && v !== null
+export const isDocument = (el: any): el is Document => el.nodeType === Node.DOCUMENT_NODE
 
-export const isHTMLElement = (el: any): el is HTMLElement =>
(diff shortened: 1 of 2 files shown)


packages/utilities/dom-query/src/is.ts
```diff
@@ -1,12 +1,18 @@
-export const isHTMLElement = (v: any): v is HTMLElement =>
-  typeof v === "object" && v?.nodeType === Node.ELEMENT_NODE && typeof v?.nodeName === "string"
+const ELEMENT_NODE: typeof Node.ELEMENT_NODE = 1
+const DOCUMENT_NODE: typeof Node.DOCUMENT_NODE = 9
+const DOCUMENT_FRAGMENT_NODE: typeof Node.DOCUMENT_FRAGMENT_NODE = 11
 
-export const isDocument = (el: any): el is Document => el.nodeType === Node.DOCUMENT_NODE
+const isObject = (v: unknown): v is Record<string, unknown> => typeof v === "object" && v !== null
```
ocavue, line 5: I know there is already a `isObject` in `@zag-js/utils` ([code link]( https://github.com/chakra-ui/zag/blob/b84a28af8e15ef5769d16d803c646ea44d6118ba/packages/utilities/core/src/guard.ts#L4 )) but that implementation also checks `Array.isArray()`, which is not necessary here.

Issue #1720, opened 2024-08-02
Clicking outside of a dialog on mobile passes click events through


Jonas-C:
# 🐛 Bug report

When clicking outside of `DialogContent` on a mobile device (or mobile emulation in the browser), click events are passed through to the actual element that is being clicked. If you have a full-screen dialog with overlapping close and open triggers in the z-axis, you're pretty much forced to reload the entire page.

1. Use a mobile device or mobile emulation in your favourite browser.
2. Open the dialog.
3. Press one of the visible (and interactive) elements still visible on the screen. 

## 💻 Link to reproduction
https://zagjs.com/components/react/dialog

## 🧐 Expected behavior

The Dialog backdrop should prevent click events from being passed down to the elements the dialog "covers"

## 🌍 System information

| Software         | Version(s) |
| ---------------- | ---------- |
| Zag Version      | 0.64.0           |
| Browser          | Firefox/Chrome/Safari           |
| Operating System |  MacOS/iOS          |

segunadebayo:
I just pushed a fix for this. We'll release an update shortly.

If the issue persists after upgrading, I'll re-open it.

PR #1743, opened 2024-08-07
fix: ensure clicks outside of dialogs/tours are not passed on to components outside of active focus trap

Jonas-C:
Fixes https://github.com/chakra-ui/zag/issues/1720

## 📝 Description

Trying my hand at fixing this, as it's currently blocking our migration to ark dialogs.

`allowOutsideClick: true` tells focus-trap that clicks on valid elements outside of the focus trap is permitted. If you have a button or link directly underneath a dialog close trigger, both the close trigger and the underlying element would be clicked.

## 🚀 New behavior

Elements outside of the dialog/trigger are no longer triggered if a dialog/tour is open.

I don't think so? I'm not really familiar with focus-trap, nor with the way zag uses it.

## 📝 Additional Information

I tried looking at the git history for both files, but there's not really any information as to why this was set in the first place. 

Also: This option is still set for popovers. I'm not really sure if you'd want popovers to ignore all outside clicks. Maybe the `modal` prop should alter it?

### DIFF
--- a/.changeset/lovely-hats-wonder.md
+++ b/.changeset/lovely-hats-wonder.md
@@ -1,0 +1,6 @@
+---
+"@zag-js/dialog": patch
+"@zag-js/tour": patch
+---
+
+fix: ensure clicks outside of dialogs are not passed on to components outside of the active focus trap
--- a/packages/machines/dialog/src/dialog.machine.ts
+++ b/packages/machines/dialog/src/dialog.machine.ts
@@ -143,7 +143,6 @@ export function machine(userContext: UserDefinedContext) {
               preventScroll: true,
               fallbackFocus: contentEl,
               returnFocusOnDeactivate: ctx.restoreFocus,
-              allowOutsideClick: true,
               initialFocus: ctx.initialFocusEl?.() ?? undefined,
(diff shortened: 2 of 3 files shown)


segunadebayo:
I appreciate the attempt @Jonas-C.

I will take a look at this over the weekend. It might have to do with using pointerdown over click for interaction outside events.

A quick fix that comes to mind is to add `onTouchEnd: e => e.preventDefault` to the content element

Jonas-C:
Is there anything stopping us from simply doing `preventDefault` and `stopPropagation` for `onTouchEnd`? See https://github.com/chakra-ui/zag/commit/c7141e8b458c5ac95b39b38a3cd6dbd2987403c8

Jonas-C:
> A quick fix that comes to mind is to add `onTouchEnd: e => e.preventDefault` to the content element

I don't think this solves the problem. I've tried adding it to `Content`, `Positioner` and `Backdrop` to no avail.

segunadebayo:
Got it. I pushed a fix for this in the main branch. This had to do with the event timings between pointerup and click on touch devices.

This issue should be gone now.

PR #1876, opened 2024-09-22
feat(qwik): Create vite plugin

anubra266:
Qwik needs to wrap event handlers and some functions in `qrls` $(...) for optimization.

This vite plugin does that at compile time.


segunadebayo:
Closing this for now. It introduces complexities we don't want to maintain.

Let's deprioritize it for now.

Issue #1912, opened 2024-10-08
integrate to Brisa framework


aralroca:
Related to this proposal https://github.com/brisa-build/brisa/issues/529 

It would be great for both (for Brisa and Chakra UI) to integrate it.

PR #1913, opened 2024-10-08
feat: add brisa adapter

anubra266:
add brisa adapter

related #1912 https://github.com/brisa-build/brisa/issues/529


anubra266:
@osdiab  @aralroca I've tried to come up with a v0 for brisa.
 There's probably more improvements that can be done on the current approach. An issue I currently face is letting the final `api.connect` be reactive. I tried using `derived` but that seems to make the whole component state/tree react when state changes. Resetting component back to initial state.
 
 Adapter is in `packages/frameworks/brisa` dir,
 While example is in `exmples/brisa` dir

aralroca:
That's cool!  🔥

aralroca:
About the behavior of JSX, it is very much inspired by React, so the difference has to be minimal. As a difference, instead of `className` is `class`, and in SVGs you don't have to do the conversion from `fill-rule`, `clip-rule` to `fillRule` or `clipRule`, already `fill-rule` or `clip-rule` (and the rest similar) work.

anubra266:
@aralroca There's a behaviour that seems problematic for the adapter, I'm keeping zag state like this:

```
 const state = state(service.state)
 ```
 service is zag's running state.
 
 When service is mutated, i seems brisa state also updates automatically. So state variable defined gets a new value. Butt, `derived` wouldn't detect this subtle change. How would I achieve something like a ref. I only want the state I defined to change when I tell it to with something like
 
 ```js
 const unsubscribe = subscribe(service.state, () => {
    state.value = snapshot(service.state)
  })
  ```

**Comment by anubra266** [comment]
Closing this for now. Will be picked up after problems are clarified

**Comment by aralroca** [comment]
I had not seen the question until now.

 In Brisa "state" also serves as "ref", and the code you mention should work, although I would not put the variable name the same. This should work:

```js
export default function Component({}, { state, onMount, cleanup }) {
  const zagState = state(service.state)

  onMount(() => {
    const unsubscribe = subscribe(service.state, () =>  {
       zagState.value = snapshot(service.state)
     })
    cleanup(() => unsubscribe())
  })

  return <div>{JSON.parse(zagState.value)}</div>
}
```

anubra266:
@aralroca Still wasn't able to get it to work. Would be awesome if you could help look into it. It's the `useSnapshot.ts` file

aralroca:
I think the problem is that in Brisa we have this condition inside the signal setter:

https://github.com/brisa-build/brisa/blob/02286643517d8a6648f064c8e34a3d28b690a341/packages/brisa/src/utils/signals/index.ts#L116

This way if the value is the same we avoid updating the DOM with the same value. I think the problem comes from here if it is mutating without being reactive inside the Zag service.

I don't know how feasible is to do something like this to force reactivity:

```js
const unsubscribe = subscribe(service.state, () => {
  // Create a new object to force reactivity even if the values are the same
  state.value = { ...snapshot(service.state) }; 
});
```

we were not contemplating that mutations could be made external to the state in this way. If this is a common behavior of several libraries, we can reconsider changing this part.

Issue #1967, opened 2024-10-22
[TagInput] onInputValueChange does not reference updated values


renchris:
# 🐛 Bug report

Any value used in the `onInputValueChange` function is not up to date despite being called after variable value change or update.

## 💥 Steps to reproduce

Update a value and see that that current value is not being used in the ``onInputValueChange` function

```ts
export const ShareCombobox = ({
  shareList,
}: {
  shareList: Share[],
}) => {
  console.log('==COMPONENT==')
  console.log(shareList) //shows here
  console.log('==')
  const handleInputChange = (details: Combobox.InputValueChangeDetails) => {
    console.log('==HANDLE INPUT CHANGE==')
    console.log(shareList) //SHOWS EMPTY LIST HERE
    console.log('==')
  }
  const handleValueChange = (details: Combobox.ValueChangeDetails) => {
    console.log('==HANDLE VALUE CHANGE==')
    console.log(shareList) //shows here
    console.log('==')
  }
//...

  const tagInput = useTagsInput({
    onValueChange: handleValueChange,
    onInputValueChange: handleInputChange,
  })
```

## 💻 Link to reproduction

CodeSandbox reproduction: https://your-codesandbox-link-goes-here.io

## 🧐 Expected behavior

I expect any variable value used in the `onInputValueChange` function to be the current value

## 🧭 Possible Solution

Update the `onInputValueChange` method in the Zag source code to update upon the function call

## 🌍 System information

| Software         | Version(s) |
| ---------------- | ---------- |
| Zag Version      |0.75.0|
| Browser          |Chrome|
| Operating System |MacOS|

segunadebayo:
I assume this is happening because the `shareList` is fetched and updated when the component mounts. This is a common stale closure problem.

For this scenario, I recommend storing the share list in a "live ref" and reading from that live ref in your code.

![CleanShot 2024-10-26 at 15 03 49@2x](https://github.com/user-attachments/assets/27489027-7208-4467-9e56-cb40a4adba89)

PR #1968, opened 2024-10-22
fix(tags input): onInputValueChange update

renchris:
Closes # 

## 📝 Description

The `onInputValueChange` function does not reference the current variable's value within the function.

The variable values used and called in the `onInputValueChange` are the old initial values and not the current values

## 🚀 New behavior

The variable values used and called in the `onInputValueChange` are the current values as expected

No

## 📝 Additional Information

Closes https://github.com/chakra-ui/zag/issues/1967

### DIFF
--- a/.changeset/friendly-trains-own.md
+++ b/.changeset/friendly-trains-own.md
@@ -1,0 +1,5 @@
+---
+"@zag-js/tags-input": patch
+---
+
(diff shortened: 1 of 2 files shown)


segunadebayo:
I tested this locally. It's not going to fix the closure problem. I recommend using the live ref approach

PR #1974, opened 2024-10-25
fix(core): ensure setTimeout delay value in machine action not overflowing

erm1116:
Closes # 

## 📝 Description
The actions function would be executed immediately if the delay of actions was specified to exceed the max range.
This PR fixes this issue by set the maximum value for setTimeout if exceeded value set.

If set the exceeded value, setTimeout's handler is executed immediately.
(strictly, the overflow results in faster execution than specified.)

## 🚀 New behavior
If set the exceeded max value, setTimeout's handler is executed after max delay time.

No

## 📝 Additional Information
`setTimeout`’s timeout value is defined as `long` type([interface](https://html.spec.whatwg.org/multipage/webappapis.html#windoworworkerglobalscope-mixin)), and `long` type’s range is [−2147483648, 2147483647] ([Web IDL](https://webidl.spec.whatwg.org/#idl-long)). 

If it is exceeded, it will be overflowing and set value as 0 in many case. ([converting algorithm](https://webidl.spec.whatwg.org/#js-long))
As a result, the handler of setTimeout will be executed immediately.

### DIFF
--- a/.changeset/little-mice-compete.md
+++ b/.changeset/little-mice-compete.md
@@ -1,0 +1,5 @@
+---
+"@zag-js/core": patch
+---
+
+Fix issue where after actions is executed immediately when setting Infinity or the over max delay value
--- a/packages/core/src/machine.ts
+++ b/packages/core/src/machine.ts
@@ -382,12 +382,18 @@ export class Machine<
   private getAfterActions = (transition: S.Transitions<TContext, TState, TEvent>, delay?: number) => {
     let id: ReturnType<typeof globalThis.setTimeout>
     const current = this.state.value!
+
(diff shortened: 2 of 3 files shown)


segunadebayo:
Thanks for this PR. I understand the intention behind this and it make sense.

I really believe this is something that should be handled outside the machine core logic.

Anyone building a machine on top of Zag should handle this use case by configuring their delays to not exceed the overflow.

erm1116:
@segunadebayo 
Thank you for your comment.
I understand that you don't want to include this kind of logic in the machine core, so I will try to handle it on the user side.

PR #1997, opened 2024-10-29
feat: improving generic component class

gperdomor:
## 📝 Description

In https://github.com/chakra-ui/zag/pull/1995 I added a custom component class to simplify Lit elements creation... Although this works fine as it's, a few changes can be done to improve types a little bit...

In current version, you can't access to code completion for `service` property.

## 🚀 New behavior

In the new version you have a proper code completion in your IDE
Screenshot 2024-10-28 at 8 46 09 PM

## 📝 Additional Information
~`MachineContext` are now exported for all machines... Exposing the `MachineState could be a good addition for further use cases~


segunadebayo:
Thanks for this PR.

I think exposing the Service is enough. We can infer the context and other things from it.

gperdomor:
@segunadebayo You're totally right... I updated the PR with a more simpler approach to archive the same results related to code completion... Let me know if you have other suggestions please...

PR #2081, opened 2024-12-04
fix: remove dpr logic due incompatibility with Firefox

### LINKED ISSUE #2080: [SignaturePad] getDataUrl seems not working properly on Firefox
# 🐛 Bug report

I was testing `getDataUrl` and on Firefox it seems not working as expected. I think it is because of `scale` transform logic we have before we generate the image.

This is the expected output, as you can see it correctly render the corners

Screenshot 2024-12-04 at 12 41 36

This is the output on Firefox. It is not the same image but you can see it only renders the half part of it

image

I did a patch in my app to remove the `win.devicePixelRatio` and always uses `1` I also changed `context!.drawImage(image, 0, 0)` to also includes width and height like this `context!.drawImage(image, 0, 0, canvas.width, canvas.height)`

hyanmandian:
Closes #2080

## 📝 Description

Remove dpr logic due incompatibility with Firefox. For some reason it seems Firefox does not work properly when we scale the canvas.

image

## 🚀 New behavior

image

### DIFF
--- a/packages/utilities/dom-query/src/data-url.ts
+++ b/packages/utilities/dom-query/src/data-url.ts
@@ -24,21 +24,19 @@ export function getDataUrl(svg: SVGElement | undefined | null, opts: DataUrlOpti
   }
 
   const svgBounds = svg.getBoundingClientRect()
-  const dpr = win.devicePixelRatio || 1
 
   const canvas = doc.createElement("canvas")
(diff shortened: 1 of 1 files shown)


segunadebayo:
Closing based on https://github.com/chakra-ui/zag/issues/2080#issuecomment-2520670509

PR #2142, opened 2024-12-29
feat: support svelte 5.16.0 class value

calvo-jp:
Closes # 

## 📝 Description

In Svelte `5.16.0`, the [class attribute accepts objects and arrays](https://github.com/sveltejs/svelte/pull/14714) as values using [clsx](https://github.com/lukeed/clsx) under the hood. This PR adds support for this feature. (code taken from `clsx`)

**Input:**

```ts
mergeProps(
  {class: {class1: true}},
  {class: ['class2', 'class3', null, undefined]},
  {class: 'class4'},
  {class: {class5: false, class6: true}},
)
```

**Output:**

```ts
{ class: 'class3' }
```

## 🚀 New behavior

**Input:**

```ts
mergeProps(
  { class: null },
  { class: undefined },
  { class: ["class1", "class2", null, undefined] },
  { class: "class3" },
  { class: { class4: false, class5: true } },
)
```

**Output:**

```ts
{ class: 'class1 class2 class3 class5' }
```

No


anubra266:
@calvo-jp It's probably better to make this change in the svelte adapter

https://github.com/chakra-ui/zag/blob/main/packages/frameworks/svelte/src/merge-props.ts

calvo-jp:
> @calvo-jp It's probably better to make this change in the svelte adapter
> 
> https://github.com/chakra-ui/zag/blob/main/packages/frameworks/svelte/src/merge-props.ts

Hi @anubra266. Will open another PR once I get back. Thank you!

segunadebayo:
After thinking deeply about this, we don't want to include this logic in zag.js since we don't use classNames in our components or logic.

Please include this in your userland or library code instead.

PR #2188, opened 2025-01-23
perf(react,preact): improve initial render

segunadebayo:
## 📝 Description

Improves the initial render times by shifting work to the framework's state primitive instead of sync-external-store logic.

## 🚀 New behavior

Machines now mount roughly 22% faster. It's a small win


segunadebayo:
Figured out a much better approach to this. It'll be released as Zag.js v1

Issue #2226, opened 2025-02-08
Timer stops when switching tabs


blasdfaa:
# 🐛 Bug report

I noticed that the timer stops when I switch to another tab and resumes when I return. Is this expected behavior? If so, what causes it?

## 💥 Steps to reproduce

1. Go to https://zagjs.com/components/react/timer
2. Remember the current timer value
3. Switch to another tab and wait some time seconds
4. Go back to timer tab
5. Timer value will be the same as before tab change.

## 💻 Link to reproduction

https://zagjs.com/components/react/timer

## 🧐 Expected behavior

Timer value must be current according to current time.

## 🧭 Possible Solution

this solution might work
https://github.com/chakra-ui/zag/compare/main...blasdfaa:zag:fix/timer-update-time

## 🌍 System information

| Software         | Version(s)            |
| ---------------- | --------------------- |
| Zag Version      | 0.82.1                |
| Browser          | Chrome 132.0.6834.160 |
| Operating System | macOS Sequoia         |

Omikorin:
This is caused by browser's behavior. Timers of running JavaScript are slowed down on inactive tabs. There are some workarounds like using web workers which are not affected by this.

blasdfaa:
> This is caused by browser's behavior. Timers of running JavaScript are slowed down on inactive tabs. There are some workarounds like using web workers which are not affected by this.

Will this be fixed at the component level?

Omikorin:
> > This is caused by browser's behavior. Timers of running JavaScript are slowed down on inactive tabs. There are some workarounds like using web workers which are not affected by this.
> 
> Will this be fixed at the component level?

It should be. Your solution is interesting, but I'll wait for Sage opinion on this machine's design. We have a few ways to fix this (and hopefully not introduce another bug)

PR #2228, opened 2025-02-09
fix(timer): timer stops when switching tabs

rewweRrr:
Closes  Timer stops when switching tabs #2226

## 📝 Description

Change calculation behaviour of `currentMs` for timer.machine

`currentMs` depends on `setInterval` it leads to incorrect calculation in case of collapsed browser tab

## 🚀 New behavior

`currentMs` calucalted independent of `interval` using only timestamp.

### DIFF
--- a/.xstate/timer.js
+++ b/.xstate/timer.js
@@ -36,6 +36,7 @@ const fetchMachine = createMachine({
       }
     },
     running: {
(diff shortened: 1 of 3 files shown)


segunadebayo:
This is now fixed in v1.

PR #2234, opened 2025-02-13
fix(hover-card): add missing type attribute to getTriggerProps

fseglard:
## 📝 Description

Add missing `type: 'button'` attribute to the `getTriggerProps` function for the `hover-card` machine.

The current `getTriggerProps` function does not return the `type: 'button'` attribute causing the HTML button element to not have the `type="button"` attribute. Without this attribute, the button's default behavior is like a "submit" button which is not the desired behavior.

## 🚀 New behavior

Now the HTML button element have the `type="button"` attribute properly added, so it won't submit the closest form on click.

No

### DIFF
--- a/.changeset/odd-mice-flash.md
+++ b/.changeset/odd-mice-flash.md
@@ -1,0 +1,5 @@
+---
+"@zag-js/hover-card": patch
+---
+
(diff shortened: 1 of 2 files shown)


segunadebayo:
HoverCard is a nuance component that might be used on a link, button, text, etc.

We left this to the userland to add the type:button as needed

PR #2237, opened 2025-02-15
feat(slider): add option "end" to the slider's origin property

### LINKED ISSUE #2236: Add option "end" to the slider's origin property next to "start" and "center"
# 🚀 Feature request

Add option `"end"` to  the slider's origin property, next to the options `"start"` and `"center"`.

The origin property of the slider sets the start position of the filled part of the track. The current options are "start" (default) and "center". However, I'm missing the option "end".

The "end" option would be useful when the value represents an offset from the end.

For example, if the slider is used to set a threshold value between 0 to 100 to include values above it, it makes more sense to have the values above the thumb highlighted with a filled track instead of the values below it.

## ✅ Proposed solution or API

I have a solution [here](https://github.com/chakra-ui/zag/compare/main...danieledler:zag:feat/slider-origin-end) and will make a pull request.

## ↩️ Alternatives

I considered using the support for right-to-left direction (dir="rtl") but I needed to wrap the slider in a locale with right-to-left direction which also has side-effects on labels.

danieledler:
Closes #2236

## 📝 Description

The slider has an origin property that sets the start position of the filled track for single values. It currently includes "start" and "center" but misses an "end" option. 

An "end" option is useful in scenarios where the slider value functions as a threshold to include values above it. Then it might make more sense to have the track filled from the thumb to the end rather than from the start to the thumb.

Now the origin property has option "start" (default) and "center". For the "start" value, the track is filled from start (left in LTR contexts, right in RTL contexts) to the thumb. For "center", the track is filled from the center (50%) to the thumb.

## 🚀 New behavior

This PR adds the origin option "end" to the slider. 

This PR also explains effect of this property more clearly.

No

### DIFF
--- a/.changeset/sixty-sides-fail.md
+++ b/.changeset/sixty-sides-fail.md
@@ -1,0 +1,5 @@
+---
+"@zag-js/slider": minor
+---
+
+Add `end` to slider `origin` property.
--- a/packages/machines/slider/src/slider.style.ts
+++ b/packages/machines/slider/src/slider.style.ts
@@ -27,6 +27,9 @@ export function getRangeOffsets(params: Pick<Ctx, "prop" | "computed">) {
       const end = isNegative ? "50%" : `${100 - valuePercent[0]}%`
       return { start, end }
     }
+    if (prop("origin") === "end") {
+      return { start: `${lastPercent}%`, end: "0%" }
+    }
 
     return { start: "0%", end: `${100 - lastPercent}%` }
   }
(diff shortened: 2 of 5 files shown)


danieledler:
### I can't commit a changelog due to missing submodule zag-plus?

After running `npx @changesets/cli` I got a file `.changeset/five-donkeys-complain.md` with the following content
```
---
"@zag-js/slider": minor
---

Add option 'end' to slider origin.
```

However, I could not commit this change. Git returns an error because it can't find the submodule `examples/next-ts/pages/compositions`.

In the [repo](https://github.com/chakra-ui/zag/tree/main/examples/next-ts/pages), compositions links to `https://github.com/chakra-ui/zag-plus` which doesn't exist. How to proceed?

danieledler:
image
Screenshot of the feature.

PR #2286, opened 2025-02-27
feat(progress): allow for more precise (decimal) values

### LINKED ISSUE #2285: [Progress] Allow decimal values
# 🚀 Feature request

I want the `progress` machine to allow for decimal values.

The progress currently [rounds to the nearest percent](https://github.com/chakra-ui/zag/blob/main/packages/machines/progress/src/progress.machine.ts#L45), I would like a way to have my progress also interpet the value as is, like the native `<progress>` element does.

## ✅ Proposed solution or API

This could be the API:
```js
    percent({ context, prop }) {
      const value = context.get("value")
      if (!isNumber(value)) return -1
      const percentage = ((value - prop("min")) / (prop("max") - prop("min"))) * 100
      return prop("allowDecimal") ? percentage : Math.round(percentage)
    },
```

## ↩️ Alternatives

I've considered making this the default behaviour and opting in with a `roundToPercentage` property, but that would be a breaking change so I'm reluctant to just implement that.

## 📝 Additional Information

The `meter` pattern doesn't require to be rounded to full numbers, so allowing decimals would still adhere to the WACG pattern: https://www.w3.org/WAI/ARIA/apg/patterns/meter/

FYI: I'm willing to make a PR

Hugos68:
Closes #2285

## 📝 Description

Removes the `Math.round` from the `percent` function to allow decimal values

If a user inputs `0.5` into the machine the output would be `1`.

## 🚀 New behavior

If a user inputs `0.5` into the machine the output would be `0.5`.

Yes, if you relied on this rounding behaviour you'd have to round it yourself like:

```diff
const service = useMachine(progress.machine, {
-   defaultValue: value
+   defaultValue: Math.round(value)
});

## 📝 Additional Information

### DIFF
--- a/packages/machines/progress/src/progress.machine.ts
+++ b/packages/machines/progress/src/progress.machine.ts
@@ -42,7 +42,7 @@ export const machine = createMachine<ProgressSchema>({
     percent({ context, prop }) {
       const value = context.get("value")
       if (!isNumber(value)) return -1
-      return Math.round(((value - prop("min")) / (prop("max") - prop("min"))) * 100)
+      return ((value - prop("min")) / (prop("max") - prop("min"))) * 100
     },
     isAtMax: ({ context, prop }) => context.get("value") === prop("max"),
     isHorizontal: ({ prop }) => prop("orientation") === "horizontal",


Hugos68:
@segunadebayo Do I add a changeset for all packages or just the `progress` machine?

segunadebayo:
Ideally, it's just for the changed package. In this case, the progress machine. @Hugos68

PR #2327, opened 2025-03-09
docs: svelte `$props.id`

Hugos68:
## 📝 Description

Recommends `$props.id` in examples

Everywhere `"1"` is used as an ID, this will obviously conflict if the examples are taken literally, Svelte now exposes `$props.id`, we should recommend users to use that (similar to `useId` in React).

## 🚀 New behavior

Adds `$props.id` to the docs

No

## 📝 Additional Information

See: https://svelte.dev/docs/svelte/$props#$props.id()


segunadebayo:
Thanks for this @Hugos68

PR #2446, opened 2025-05-09
Add OKLAB and OKLCH to color-picker

V1OL3TF0X:
Addresses [@2439](https://github.com/chakra-ui/zag/discussions/2439)

## 📝 Description

This PR adds two new modes to the color-picker: oklch and oklab. I also needed to fix a bit of the logic, so switching between modes worked properly. A also tried to update the examples to the best of my abilities

- color pikcer has 3 modes: RGBA, HSLA and HSBA

## 🚀 New behavior

- color-picker has 2 new modes - OKLAB and OKLCH.
- examples are updated to see effects of switching modes, as well as show all channels of the current mode under area picker
- switching the mode doesn't break the UI

No

## 📝 Additional Information

If anything falls outside contribution guidelines, please let me know - I tried to follow the code style to the best of my abilities, but some things may have slipped through.

As parsing oklab and oklch has some complicated math involved, I opted to add [culori/js](https://culorijs.org) to take care of that


segunadebayo:
@V1OL3TF0X thanks for this PR. Love the direction.

Could we do this without an additional dependency? We want to keep the bundle size roughly similar

V1OL3TF0X:
> @V1OL3TF0X thanks for this PR. Love the direction.
> 
> Could we do this without an additional dependency? We want to keep the bundle size roughly similar

Done - I leaned out the dependency as much as possible. This has the downside of not clamping by chroma but by RGB (less accurate, but I'm not smart enough to translate `clampChroma` from culori). I decided to keep the conversion functions, since rewriting them wouldn't impact the bundle size as they're dependency free, and could only lead to some mistakes during the process.

I also updated the docs a bit, to include the new modes

V1OL3TF0X:
@segunadebayo what's the status on this? Should I add the version bump, is the docs update good? If not, what's there to change?

Hugos68:
@segunadebayo What's the status on this? We really want to use the Color picker but OKLCH support is a must for us.

V1OL3TF0X:
@segunadebayo How's the review process going?

V1OL3TF0X:
@segunadebayo any updates on this?

segunadebayo:
I've added support for this in the upcoming v2 branch. We'll publish it with v2

PR #2500, opened 2025-06-03
feat: add cascade select machine

anubra266:
Add cascader


segunadebayo:
Tested this out on mobile and clicking a nested item doesn’t seem to work.

PR #2509, opened 2025-06-08
fix(carousel): Prevent infinite loop when slidesPerPage is 0

### LINKED ISSUE #2508: React Carousel: Documentation page crashes when slidesPerView is set to 0
🐛 Bug Report

On the React Carousel documentation page, modifying the slidesPerView control to an invalid value (specifically 0) causes the entire page to become unresponsive and crash. The browser tab freezes, and the developer console is flooded with "Maximum call stack size exceeded" errors, making the component and the page unusable until it's refreshed.

To Reproduce
Go to the React Carousel documentation page: https://zagjs.com/components/react/carousel
On the right-hand side, locate the Controls panel.

Find the number input for slidesPerView. It defaults to 1.
Change the value to 0 (either by typing 0 or using the stepper's down arrow).

Observe the page immediately freeze and become unresponsive.

Expected behavior

The component or the input control should handle invalid values gracefully. Some expected behaviors could be:
The input field should have a minimum value validation (e.g., min="1") to prevent users from entering 0 or negative numbers.

If an invalid value is entered, the component should not crash. It could either ignore the invalid value, fall back to a default (like 1), or show a validation message.
Actual behavior

The page crashes.

miladbn:
This pull request resolves the high-priority bug reported in #2508,

 where the carousel machine enters an infinite loop if the slidesPerPage prop is set to 0.

**The Problem**
As identified in the issue, when slidesPerPage is 0 and slidesPerMove is "auto", the perMove variable within the getPageSnapPoints utility is calculated as 0. This causes the for loop responsible for generating snap points to become an infinite loop (i += 0), leading to the browser tab hanging and crashing.

The Solution
This PR implements a simple guard clause at the beginning of the getPageSnapPoints function.
It now checks if slidesPerPage is 0 before any calculations occur.
If the condition is met, the function immediately returns an empty array, gracefully handling the invalid configuration and preventing the infinite loop.

This ensures the application remains stable even with this edge-case configuration.

How to Verify
To Reproduce the Bug (on main):
Render the carousel component with the prop slidesPerPage={0}.

The browser tab will freeze.
To Confirm the Fix (on this branch):
Render the carousel component with slidesPerPage={0}.
The application will run without errors, and the page remains responsive.
Related Issues

Closes #2508
Checklist
My code follows the project's style guidelines.
I have tested my changes locally and they resolve the reported bug.
The commit message is clear and follows conventional commit standards.
I have added corresponding tests for my changes (if applicable).
I have updated the documentation accordingly (if applicable).

### DIFF
--- a/packages/machines/carousel/src/carousel.machine.ts
+++ b/packages/machines/carousel/src/carousel.machine.ts
@@ -481,9 +481,21 @@ export const machine = createMachine<CarouselSchema>({
 })
 
 function getPageSnapPoints(totalSlides: number | undefined, slidesPerMove: number | "auto", slidesPerPage: number) {
-  if (totalSlides == null) return []
+  // If there are no total slides or if slidesPerPage is 0, do nothing.
+  if (totalSlides == null || slidesPerPage === 0) {
+    return []
+  }
+
   const snapPoints: number[] = []
   const perMove = slidesPerMove === "auto" ? Math.floor(slidesPerPage) : slidesPerMove
-  for (let i = 0; i < totalSlides - 1; i += perMove) snapPoints.push(i)
+
+  // An additional safety check in case perMove becomes 0 through other means
+  if (perMove === 0) {
+    return []
+  }
+
+  for (let i = 0; i < totalSlides - 1; i += perMove) {
+    snapPoints.push(i)
+  }
   return snapPoints
 }
--- a/packages/machines/carousel/src/carousel.machine.ts
+++ b/packages/machines/carousel/src/carousel.machine.ts
@@ -482,7 +482,7 @@ export const machine = createMachine<CarouselSchema>({
 
 function getPageSnapPoints(totalSlides: number | undefined, slidesPerMove: number | "auto", slidesPerPage: number) {
   // If there are no total slides or if slidesPerPage is 0, do nothing.
-  if (totalSlides == null || slidesPerPage === 0) {
+  if (totalSlides == null || slidesPerPage <= 0) {
     return []
   }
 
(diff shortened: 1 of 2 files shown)


packages/machines/carousel/src/carousel.machine.ts
```diff
@@ -481,9 +481,21 @@ export const machine = createMachine<CarouselSchema>({
 })
 
 function getPageSnapPoints(totalSlides: number | undefined, slidesPerMove: number | "auto", slidesPerPage: number) {
-  if (totalSlides == null) return []
+  // If there are no total slides or if slidesPerPage is 0, do nothing.
+  if (totalSlides == null || slidesPerPage === 0) {
```
segunadebayo, line 485:
```suggestion
  if (totalSlides == null || slidesPerPage <= 0) {
```

packages/machines/carousel/src/carousel.machine.ts
```diff
@@ -481,9 +481,21 @@ export const machine = createMachine<CarouselSchema>({
...
+  if (totalSlides == null || slidesPerPage <= 0) {
+    return []
+  }
+
   const snapPoints: number[] = []
   const perMove = slidesPerMove === "auto" ? Math.floor(slidesPerPage) : slidesPerMove
-  for (let i = 0; i < totalSlides - 1; i += perMove) snapPoints.push(i)
+
+  // An additional safety check in case perMove becomes 0 through other means
+  if (perMove === 0) {
```
segunadebayo, line 493:
```suggestion
  if (perMove <= 0) {
```

PR #2524, opened 2025-06-18
fix(number-input): activate scrubber only on main button

kristian240:
## 📝 Description  & ⛳️ Current behavior (updates)

When you right click on scrubber icon the conext menu is opened but cursor cannot be moved. The issue is not present in Safari since they do not support `requestPointerLock` API

## 🚀 New behavior

This PR changes the behaviour to only activate a scrubber on primary button click. Uses [MouseEvent: button property](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button) to detect that.

No


segunadebayo:
Thanks for this PR @kristian240.

I've extended this to other components as well in the main branch.

PR #2562, opened 2025-07-01
fix(progress): improve `valueAsString` formatting

### LINKED ISSUE #2560: Circular Progress value is not formatted properly
# 🐛 Bug report

I'm reporting this issue on behalf of [Skeleton](https://www.skeleton.dev/), you can [read through the context here](https://github.com/skeletonlabs/skeleton/issues/3539#issuecomment-2887306731).

> FYI: Skeleton refers to the Circular Progress as "Progress Ring" for consistency on our end.

Essentially it's not clear if the Circular Progress should be auto-formatting the `valueAsString` to match the current `formatOptions` settings.

As it stands, the default value for this setting is as `percent`:

Image

And it should support these three other optional `style` settings:

Image

Currently this only seems to work with the default (percent) as follows:

- `value` -> shows the raw numeric value, such as `50`
- `valueAsString` -> shows a formatted string, such as `50 Percent`

So I would expect similar changes for `decimal`, `currency`, and `unit`. But this does not appear to be the case. When setting `decimal` for example, the `valueAsString` still returns as `50 Percent`.

## 💥 Steps to reproduce

1. Scaffold the Circular Progress Zag component as shown in the docs
2. Pass a value (ex: 50)
3. Set `formatOptions` as `{ style: 'percent' }` (the default setting)
4. The results will show "50 Percent" as expected.
5. Now change value to `1.5` and `formatOptions` to `{ style: 'decimal' }`
6. This will show the string `1.5 Percent` instead of the expected `1.5`

So the formatter seems to always use the `percent` format.

## 💻 Link to reproduction

Here's a barebones React boilerplate to test:

```tsx
import * as progress from "@zag-js/progress";
import { normalizeProps, useMachine } from "@zag-js/react";
import { useId } from "react";

function App() {
  const service = useMachine(progress.machine, {
    id: useId(),
    value: 50,
    formatOptions: { style: "percent" },
  });

  const api = progress.connect(service, normalizeProps);

  return (
    <div {...api.getRootProps()}>
      {/* <div {...api.getLabelProps()}>LabelHere</div> */}
      {/* <svg {...api.getCircleProps()}>
        <circle {...api.getCircleTrackProps()} />
        <circle {...api.getCircleRangeProps()} />
      </svg> */}
      <pre>value: {api.value}</pre>
      <pre>valueAsString: {api.valueAsString}</pre>
    </div>
  );
}

export default App;
```

Which will display as follows:

Image

## 🧐 Expected behavior

Again, modifying `formatOptions.style` will have no affect on either the `value` or `valueAsString` per expectations. Try any of the following:

* `formatOptions: { style: "decimal" }`
* `formatOptions: { style: "currency" }`
* `formatOptions: { style: "unit" }`

All three options will match display in the screenshot above.

## 🧭 Possible Solution

Idea 1:

1. I'd assume `api.value` should always be a verbatim copy of `value`
2. `api.valueAsString` should adhere to the `formatOptions` style
3. I might also suggest supplementary options to further refine the final display. Such as: 

```
formatOptions: { style: "percent", symbol: "%" } // result -> `50%`
formatOptions: { style: "currency", format: "USD" } // result -> `$50.00`
```

Idea 2:

Alternatively, if Zag doesn't want to be responsible for formatting options, it would be nice to capture this with a `formatOptions` prop, but allow us (Skeleton) to read this setting and implement our own local formatter. In our case, we would NEVER want the word `Percent` over `%` as the text string is not i18n friendly.

## 🌍 System information

| Zag Version      | 1.18.0 |
| Browser          | Chrome 138.0.7204.93 |
| Operating System | macOS Sequoia 15.5 |

## 📝 Additional information

In summary there's three parts to this:

1. Clarifying the expectations for the `formatOptions` setting
2. Assuming it's meant to auto-format, treating this as a bug as this only works 1 of 4 options
3. Assuming it's not meant to auto-format, exposing a means easily read `formatOptions`

Thanks!

anubra266:
Closes #2560

image
image


endigo9740:
Amazing work, thanks @anubra266 and @segunadebayo

PR #2581, opened 2025-07-17
fix(date-picker): prefer date time contraint to allow dates set to the end of day

maastrich:
## 📝 Description

Fix date picker constraint logic to properly handle end-of-day times by using `toCalendarDateTime` instead of `toCalendarDate` when applying min/max constraints.

When constraining dates to min/max boundaries, the date picker was using `toCalendarDate()` which strips time information, forcing dates to the beginning of the day (e.g., `2023-12-31T00:00:00`). This prevented users from setting dates to the end of the specified day range.

## 🚀 New behavior

The date picker now uses `toCalendarDateTime()` to preserve time information when applying constraints, allowing dates to be set to the end of the day (e.g., `2023-12-31T23:59:59.999`). This provides more accurate and expected constraint behavior for date ranges.

## 💣 Is this a breaking change (Yes/No): No

This is a bug fix that improves constraint accuracy without changing the public API.

## 📝 Additional Information

- Updated `constrainValue()` and `constrainStart()` functions in date-utils
- Enhanced test coverage to verify end-of-day constraint behavior
- Affects date picker min/max value handling across all framework adapters

### DIFF
--- a/.changeset/polite-cobras-pull.md
+++ b/.changeset/polite-cobras-pull.md
@@ -1,0 +1,5 @@
+---
+"@zag-js/date-utils": patch
+---
+
+Fix date picker constraint logic to properly handle end-of-day times by using `toCalendarDateTime` instead of `toCalendarDate` when applying min/max constraints.
--- a/packages/utilities/date-utils/src/constrain.ts
+++ b/packages/utilities/date-utils/src/constrain.ts
@@ -6,7 +6,7 @@ import {
   startOfMonth,
   startOfWeek,
   startOfYear,
-  toCalendarDate,
+  toCalendarDateTime,
 } from "@internationalized/date"
 
 /* -----------------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


segunadebayo:
CleanShot 2025-07-18 at 12 25 26@2x

I wonder if this change can be considered a breaking change. If a user already relies on that dateString, it might affect them.

If there's a way to strip out the time component after the constraint, then it should be fine.

maastrich:
> CleanShot 2025-07-18 at 12 25 26@2x
> I wonder if this change can be considered a breaking change. If a user already relies on that dateString, it might affect them.
> 
> If there's a way to strip out the time component after the constraint, then it should be fine.

Hey @segunadebayo, I see what you mean but I'm thinking the user should be in control of the format it should match against 👀 
what do you think about the latest commit ?
https://github.com/chakra-ui/zag/pull/2581/commits/08364fd76c7699d5949af82117f2e7f2c6bf0390

segunadebayo:
I've refactored the utils only to use the date component and ignore the time components. We don't have support for time components in the datepicker, when we do, we'll improve it.

PR #2583, opened 2025-07-17
feat: add Hono JSX wrapper

3w36zj6:
## 📝 Description

This PR adds a wrapper for [Hono JSX](https://hono.dev/docs/guides/jsx). I believe Hono is a very attractive library, but as far as I know, there are no component libraries that support Hono JSX. Making integration with zag easier will be beneficial for users of both libraries.

## 🚀 New behavior

Hono users can now become zaggers by using `@zag-js/hono-jsx`.

No.

### DIFF
--- a/examples/hono-jsx-ts/app/routes/accordion/$Accordion.tsx
+++ b/examples/hono-jsx-ts/app/routes/accordion/$Accordion.tsx
@@ -1,0 +1,49 @@
+import * as accordion from "@zag-js/accordion"
+import { normalizeProps, useMachine } from "@zag-js/hono-jsx"
+import { useId } from "hono/jsx"
+
+const items = [
+  {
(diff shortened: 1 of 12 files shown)


segunadebayo:
Nice. Never thought about this. It’s an interesting use case for Zag.

Could you create a single component example so we can test it out?

3w36zj6:
Hi @segunadebayo,

I've created an example project in `examples/hono-jsx-ts` for testing the Hono JSX wrapper.
Let me know if you need anything else or further examples.

segunadebayo:
After much consideration, I'd prefer we keep this a community project instead of baking it into zag.js

Just like this Ripple.js support: https://github.com/anubra266/zag-ripple

Feel free to create a dedicated repo for the project and share it with the world.

Thanks for understanding

PR #2591, opened 2025-07-21
fix(date): improve range

anubra266:
Consider `hoveredRangeValue` for day cell state in `firstInRange` and `lastInRange`.
This allows for targetting first and last hovered items.

### DIFF
--- a/.changeset/hip-spoons-design.md
+++ b/.changeset/hip-spoons-design.md
@@ -1,0 +1,6 @@
+---
+"@zag-js/date-picker": patch
+---
+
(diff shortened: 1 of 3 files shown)


segunadebayo:
After thinking about it, this one is incorrect. 

We only want to add firstInRange and lastInRange for selection, not hovering. Added it to isInRange just for styling the items in range. Until a selection is made, we don't want to add those data-attr.

A better approach might be to include an attribute when the user is attempting to select a range.

segunadebayo:
I've implemented a different version of this in the main branch

PR #2593, opened 2025-07-21
fix(date-picker): normalize dates for accurate comparisons

anubra266:
Normalize dates for accurate comparisons.
Strip time data, so date comparisons are accurate


segunadebayo:
Fixed this and added tests in the main branch

PR #2608, opened 2025-07-31
feat: added lit examples... again

gperdomor:
Closes # 

## 📝 Description

This PR add Lit examples

There is not Lit examples provided

## 🚀 New behavior

Adds a POC of using zag with Lit

No


manuelserradev:
Sorry to interrupt — outsider here 👋

I noticed some back-and-forth on this PR regarding the Lit example. Just a thought: aligning the Lit utilities with the semantics used in other frameworks might help ease the merge.

For instance, proxying `new VanillaMachine(...props)` to a hook-like API such as `useMachine(...props)` could bring the Lit example closer to the official ones. That might even pave the way for Lit to become officially supported, especially if a few champions step up.

This could also broaden Zag's scope as a Swiss Army knife for building highly accessible, cross-framework design systems.

Disclaimer: apologies if any of this is off-base — I'm not an expert in Lit or Zag, just sharing an outside perspective.

PR #2609, opened 2025-08-02
feat(tab | radio-group) indicator disabled

isBatak:
Closes # 

## 📝 Description

> Add a brief description

## 🚀 New behavior


segunadebayo:
After considering this further, I don't think we should support this. 

There are alternative ways to work around this feature:
- css targeting (e.g., `:has([data-selected]:disabled) .indicator`).
- using `api.getItemState({ value: tabs.value })`, then checking if the active item is disabled and applying some styles to the indicator

The latter approach could be the solution we're looking for. Not sure

> I apologize for leading you to work on this.

PR #2630, opened 2025-08-13
feat(radio-group): maintain data-focus-visible on arrow key navigation

### LINKED ISSUE #2629: [Radio Group] data-focus-visible not set when navigating with arrow keys
# 🐛 Bug report

When using @zag-js/radio-group (Ark UI <RadioGroup>), the data-focus-visible attribute is only applied when the group first receives focus via keyboard (e.g. Tab).
When navigating between items with arrow keys, data-focus moves correctly, but data-focus-visible is not applied to the newly focused item.
This is inconsistent with the [WAI-ARIA Authoring Practices 1.2 – Radio Group Keyboard Interaction](https://www.w3.org/WAI/ARIA/apg/patterns/radio/), where focus should visibly follow the selected radio during arrow key navigation.

## 💥 Steps to reproduce

1. Go to the [Ark UI Radio Group docs example](https://ark-ui.com/docs/components/radio-group).
2. Use the Tab key to focus the first radio item — data-focus-visible is applied.
3. Use the arrow keys (↑/↓/←/→) to move to another radio item.
4. Observe that data-focus moves to the new item, but data-focus-visible is absent.

## 🧐 Expected behavior

- When navigating between radio buttons via arrow keys, the newly focused item should also have data-focus-visible applied, so that focus indicators remain visible for keyboard users.
- This matches the behavior of native :focus-visible, which shows the focus ring during arrow key navigation inside radio groups.

## 🧭 Possible Solution

- Update the @zag-js/focus-visible modality detection so that arrow key–initiated focus changes within the same widget also set data-focus-visible.
- Align behavior with the native :focus-visible pseudo-class, which handles arrow key navigation correctly.

## 🌍 System information

| Zag Version      |      latest      |
| Browser          |      Chrome      |
| Operating System |      MacOS      |

devpla:
Closes #2629

## 📝 Description

Ensures that the data-focus-visible attribute is preserved when navigating between items in the RadioGroup component using arrow keys.
This improves keyboard accessibility by keeping the visual focus indicator visible throughout arrow key navigation, aligning with WAI-ARIA Authoring Practices.

Currently, data-focus-visible is only applied when the RadioGroup first receives focus via keyboard entry (e.g., Tab).
When navigating with arrow keys, focusedValue updates to the new item, but data-focus-visible is not applied, causing the visual focus indicator to disappear for keyboard users.

## 🚀 New behavior

- Added focusVisibleValue state and SET_FOCUS_VISIBLE action
- Updated keyboard event handling to update focusVisibleValue when navigating via arrow keys
- Changed data-focus-visible logic to use focusVisibleValue instead of focusedValue
- As a result, data-focus-visible is maintained during both initial Tab entry and arrow key navigation

No

## 📝 Additional Information

- This change follows the [WAI-ARIA Authoring Practices 1.2 – Radio Group Keyboard Interaction](https://www.w3.org/WAI/ARIA/apg/patterns/radio/) guidelines.
- focusedValue still tracks the focus position, while focusVisibleValue specifically tracks whether the focus should be visually indicated based on keyboard modality.
- If this approach differs from the intended design or there’s a better integration path, feedback would be greatly appreciated.

### DIFF
--- a/packages/machines/radio-group/src/radio-group.connect.ts
+++ b/packages/machines/radio-group/src/radio-group.connect.ts
@@ -16,7 +16,7 @@ export function connect<T extends PropTypes>(
 
   function getItemState(props: ItemProps): ItemState {
     const focused = context.get("focusedValue") === props.value
-    const focusVisible = focused && isFocusVisible()
+    const focusVisible = context.get("focusVisibleValue") === props.value
     return {
       value: props.value,
       invalid: !!props.invalid,
@@ -172,9 +172,14 @@ export function connect<T extends PropTypes>(
         },
         onBlur() {
           send({ type: "SET_FOCUSED", value: null, focused: false })
+          send({ type: "SET_FOCUS_VISIBLE", value: null, focusVisible: false })
         },
         onFocus() {
           send({ type: "SET_FOCUSED", value: props.value, focused: true })
+          const focusVisible = isFocusVisible()
+          if (focusVisible) {
+            send({ type: "SET_FOCUS_VISIBLE", value: props.value, focusVisible: true })
+          }
         },
         onKeyDown(event) {
           if (event.defaultPrevented) return
--- a/packages/machines/radio-group/src/radio-group.machine.ts
+++ b/packages/machines/radio-group/src/radio-group.machine.ts
@@ -34,6 +34,9 @@ export const machine = createMachine<RadioGroupSchema>({
       focusedValue: bindable<string | null>(() => ({
         defaultValue: null,
       })),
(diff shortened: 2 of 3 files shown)


devpla:
@segunadebayo Thanks for the update!
I noticed that this change has already been addressed in [ad58108](https://github.com/chakra-ui/zag/commit/ad58108780806a741bfba7a172c600fda4e770d3), which caused a merge conflict.
Since the fix is now covered by that commit, I’ll close this PR to avoid duplication.

segunadebayo:
Thanks @devpla.

I appreciate you attempting to fix this.

PR #2668, opened 2025-08-26
feat: sync trigger readonly implementation between date and time picker machine

isBatak:
## 📝 Description

 sync trigger readonly implementation between date and time picker machine

## 🚀 New behavior


segunadebayo:
Removed time-picker for now. It's poorly designed and doesn't have the time to dedicate to it.

For now, I suggest we use a time grid or even a time select as an alternative, which offers a simpler UX.

PR #2695, opened 2025-09-09
fix(menu): merge props before normalizing in getTriggerItemProps

nikparo:
Closes # 

## 📝 Description

> Add a brief description

The menu `getItemProps` and `getTriggerProps` return normalized props, that then get merged using the core `mergeProps` function that does not take normalization into account. This is a problem for e.g. Lit, where the event handler prefix is `@` rather than `on`. To properly merge these sets of props in the `connect` function we need to separate handling of framework-specific and framework-agnostic prop handling.

If the `normalizeProps` function does not use `on` prefix for event handlers, only the latter is kept.

## 🚀 New behavior

Event handlers are properly merged, as prop normalization happens after the merge.

No

### DIFF
--- a/packages/machines/menu/src/menu.connect.ts
+++ b/packages/machines/menu/src/menu.connect.ts
@@ -1,5 +1,6 @@
 import type { Service } from "@zag-js/core"
 import { mergeProps } from "@zag-js/core"
+import { createNormalizer } from "@zag-js/types"
 import {
   ariaAttr,
   dataAttr,
@@ -22,6 +23,9 @@ import { cast, hasProp } from "@zag-js/utils"
 import { parts } from "./menu.anatomy"
 import * as dom from "./menu.dom"
 import type { ItemProps, ItemState, MenuApi, MenuSchema, OptionItemProps, OptionItemState } from "./menu.types"
+
+// We need to avoid mixing framework-agnostic logic with framework-specific prop handling
+const identityProps = createNormalizer<PropTypes>((v) => v)
 
 export function connect<T extends PropTypes>(service: Service<MenuSchema>, normalize: NormalizeProps<T>): MenuApi<T> {
   const { context, send, state, computed, prop, scope } = service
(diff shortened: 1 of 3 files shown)


segunadebayo:
Hey @nikparo, thanks for the detailed write-up — the bug you've identified is real, but I think the fix belongs in your Lit adapter PR (#2698) rather than in the menu connect logic.

Here's why: for all currently supported frameworks (React, Solid, Vue, Svelte, Preact, Vanilla), `normalizeProps` preserves the `on` prefix for event handlers, so core `mergeProps` chains them correctly. The issue only surfaces when a framework normalizer changes the event handler prefix (e.g., `onClick` → `@click` in Lit).

Instead of changing the menu internals, I'd suggest providing a custom `mergeProps` in `@zag-js/lit` that bridges the `@` ↔ `on` prefix gap:

```ts
// packages/frameworks/lit/src/merge-props.ts
import { mergeProps as coreMergeProps } from "@zag-js/core"

export function mergeProps<T extends Record<string, any>>(
  ...args: Array<T | undefined>
): T {
  // Convert @-prefixed event handlers to on- prefix
  // so core mergeProps chains them correctly, then convert back.
  const normalized = args.map((props) => {
    if (!props) return props
    const result: Record<string, any> = {}
    for (const key in props) {
      if (key.startsWith("@")) {
        result[`on${key.slice(1)}`] = props[key]
      } else {
        result[key] = props[key]
      }
    }
    return result as T
  })

  const merged = coreMergeProps(...normalized)

  // Convert back to @-prefixed
  const result: Record<string, any> = {}
  for (const key in merged) {
    if (key.startsWith("on") && typeof merged[key] === "function") {
      result[`@${key.slice(2).toLowerCase()}`] = merged[key]
    } else {
      result[key] = merged[key]
    }
  }

  return result as T
}
```

This keeps the framework-specific concern in the framework adapter, which is the whole point of the adapter pattern. Every other framework adapter (React, Solid, etc.) re-exports `mergeProps` from core — Lit would just provide its own.

Let me know if you run into issues with this approach in #2698!

segunadebayo:
Closing — see comment above. The fix belongs in the Lit adapter PR (#2698).

PR #2697, opened 2025-09-09
fix: account for shadow-dom in activeElement checks

nikparo:
Closes # 

## 📝 Description

> Add a brief description

`activeElement` checks are sometimes done using `document.activeElement !== element`, which is an issue for apps containing shadow-dom elements such as web-components. This prevents Zag from being used in e.g. default Lit setups (https://github.com/chakra-ui/zag/discussions/2681).

- Add `isActiveElement` to `@zag-js/dom-query` for consistent activeElement checks.
- `getActiveElement` is tweaked to allow for focusable (`tabindex="0"`) web components that do not contain 

Focus trap and other utilities / machines do not find the focused elment within a shadow-dom root, resulting in unwanted behaviour.

## 🚀 New behavior

Focused elements are consistently found.

No

### DIFF
--- a/packages/utilities/dom-query/src/node.ts
+++ b/packages/utilities/dom-query/src/node.ts
@@ -104,7 +104,7 @@ export function getActiveElement(rootNode: Document | ShadowRoot): HTMLElement |
   let activeElement = rootNode.activeElement as HTMLElement | null
   while (activeElement?.shadowRoot) {
     const el = activeElement.shadowRoot.activeElement as HTMLElement | null
-    if (el === activeElement) break
+    if (!el || el === activeElement) break
     else activeElement = el
   }
   return activeElement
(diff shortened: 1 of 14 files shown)


packages/utilities/dom-query/src/node.ts
```diff
@@ -37,6 +37,13 @@ export const isElementVisible = (el: Node) => {
   return el.offsetWidth > 0 || el.offsetHeight > 0 || el.getClientRects().length > 0
 }
 
+export function isActiveElement(element: Element | null | undefined): boolean {
+  if (!element) return false
+  const rootNode = element.getRootNode() as Document | ShadowRoot
+
+  return getActiveElement(rootNode) === element
+}
```
nikparo, line 45: I'm unsure if this function should be used in the `createScope` functions or not. Currently those check both whether the element is active and within the scope root. Is that intentional? It seems to me like that would be fragile e.g. if the component uses some kind of portals or renders something outside the scope root.

nikparo: Like this you mean? 8c8537da7445411c36b2c8a63394d57156f242c9 + a51a77c5ff347e128a24f14b7732d5d6e44e8b0b
With this `isActiveElement` is only used directly in `dom-query` and `core`.

PR #2723, opened 2025-09-18
formatByte: Add unit system property

IvanKalinin:
In computing, there are two competing definitions of what “kilo,” “mega,” “giga,” etc., mean when applied to bytes:

**Decimal** (SI definition)
1 kilobyte (KB) = 1000 bytes
1 megabyte (MB) = 1000² bytes
Common in disk storage, networking, and marketing.

**Binary** (IEC definition)

1 kibibyte (KiB) = 1024 bytes
1 mebibyte (MiB) = 1024² bytes
Used in memory, OS-level reporting, and low-level computing

I added `definition` so developers can choose between default `decimal` (1000) and `binary` (1024) unit systems, ensuring accurate and context-appropriate byte formatting.

### DIFF
--- a/.changeset/little-eyes-dress.md
+++ b/.changeset/little-eyes-dress.md
@@ -1,0 +1,6 @@
+---
+"@zag-js/i18n-utils": patch
+---
+
+- **formatBytes**: Support `unitSystem` property to allow changing between decimal (1000 bytes) and binary (1024 bytes)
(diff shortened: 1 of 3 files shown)


segunadebayo:
That's a good addition, @IvanKalinin.

I've renamed the property slightly to clarify its intent better

PR #2746, opened 2025-10-02
fix(listbox): update `splitProps` to avoid partial

### LINKED ISSUE #2745: [`@zag-js/listbox`] `splitProps` returns Partial<Props> instead of Props
# 🐛 Bug report

The `splitProps` function from the listbox machine returns a partial props tuple instead of a props tuple. 

## 💥 Steps to reproduce

1. Use `listbox.splitProps`
2. View the tuple returned be `Partial`

## 💻 Link to reproduction

CodeSandbox reproduction: https://your-codesandbox-link-goes-here.io

## 🧐 Expected behavior

It should return `Props` not `Partial<Props>`

## 🌍 System information

| Zag Version      | 1.25.0 |

## 📝 Additional information

Image

Current workaround is aserrting the type to:
```ts
as unknown as [Props, Omit<ListboxRootProps, keyof Props>];
```

julienbenac:
Closes #2745

## 📝 Description

This PR fixes a TypeScript typing issue in the `Listbox` machine where the `splitProps` function incorrectly returns `Partial<ListboxProps>` instead of `ListboxProps`. This change corrects the type parameter to ensure proper type inference and eliminates the need for manual type assertions.

The `splitProps` function was defined with `Partial<ListboxProps>` as the type parameter, which made all properties optional in the returned props object. This caused TypeScript typing issues where users had to use workaround type assertions.

## 🚀 New behavior

The `splitProps` function now correctly returns typed props where required properties remain required and optional properties remain optional, matching the original `ListboxProps` interface definition.

No.

## 📝 Additional Information

This fix aligns the `Listbox` machine with the correct usage pattern found in other machines and matches the intended behavior of the `createSplitProps` utility function.

#### Before

```typescript
// ❌ Using Partial<ListboxProps> makes all properties optional
export const splitProps = createSplitProps<Partial<ListboxProps>>(props)
```

#### After

```typescript
// ✅ Using ListboxProps preserves required/optional property types
export const splitProps = createSplitProps<ListboxProps>(props)
```

### DIFF
--- a/packages/machines/listbox/src/listbox.props.ts
+++ b/packages/machines/listbox/src/listbox.props.ts
@@ -25,7 +25,7 @@ export const props = createProps<ListboxProps>()([
   "typeahead",
   "value",
 ])
-export const splitProps = createSplitProps<Partial<ListboxProps>>(props)
+export const splitProps = createSplitProps<ListboxProps>(props)
 
 export const itemProps = createProps<ItemProps>()(["item", "highlightOnHover"])
 export const splitItemProps = createSplitProps<ItemProps>(itemProps)


segunadebayo:
Thanks. Please add a changeset and resolve any TS errors

PR #2787, opened 2025-10-17
fix(select): adds missing aria-required to trigger props

### LINKED ISSUE #2785: Required property on Select component not reflected in accessibility tree
# 🐛 Bug report

When using the required property on the custom Select component from Zag.js, it has no effect on the accessibility tree. Assistive technologies do not detect the component as required, which may lead to accessibility issues for users relying on screen readers or other tools.

This suggests that the required state is either not properly set on the underlying DOM element or not exposed via ARIA attributes, resulting in a missing announcement of the required status.

## 💥 Steps to reproduce

1. Go to the page where the Zag.js custom Select component is rendered
2. Ensure the required property is set on the component
3. Open browser DevTools and inspect the accessibility tree
4. Observe that the Select component is not marked as required, despite the required property being set

## 💻 Link to reproduction

CodeSandbox reproduction: https://stackblitz.com/edit/pqeavtev?file=src%2FApp.tsx

## 🧐 Expected behavior

The Select component should be exposed as required in the accessibility tree. Assistive technologies should be able to detect and announce the required state, either via native HTML attributes or appropriate ARIA properties.

## 🧭 Possible Solution

Update the getTriggerProps function in the Zag.js Select component to include aria-required="true" when the required property is set—similar to how aria-invalid is currently handled.

## 🌍 System information

| Software | Version(s) |
| Zag Version | 1.26.2 |
| Browser | Chrome 141.0.7390.108  |
| Operating System | macOS Tahoe 26.0.1  |

## 📝 Additional information

Image

Hagendorn:
Closes #2785

## 📝 Description

This PR adds accessibility support for the required state in the Zag.js Select component by ensuring it is reflected in the accessibility tree.

Currently, when the required property is set on the Select component, it has no effect on the accessibility tree. Assistive technologies do not detect the component as required, which can lead to accessibility issues and non-compliance with WCAG guidelines.

## 🚀 New behavior

The getTriggerProps function now sets aria-required="true" when the required property is present. This ensures that the required state is properly exposed to assistive technologies, similar to how aria-invalid is handled.

No

## 📝 Additional Information
Bildschirmfoto 2025-10-17 um 16 32 17
Bildschirmfoto 2025-10-17 um 16 32 22

### DIFF
--- a/.changeset/tidy-dingos-deny.md
+++ b/.changeset/tidy-dingos-deny.md
@@ -1,0 +1,5 @@
+---
+"@zag-js/select": patch
+---
+
+Fix accessibility violation where the required state was not set correctly to on the trigger.
--- a/packages/machines/select/src/select.connect.ts
+++ b/packages/machines/select/src/select.connect.ts
@@ -171,6 +171,7 @@ export function connect<T extends PropTypes, V extends CollectionItem = Collecti
         "aria-haspopup": "listbox",
         "data-state": open ? "open" : "closed",
         "aria-invalid": invalid,
+        "aria-required": required,
         "aria-labelledby": dom.getLabelId(scope),
         ...parts.trigger.attrs,
         "data-disabled": dataAttr(disabled),
--- a/shared/src/controls.ts
+++ b/shared/src/controls.ts
@@ -166,6 +166,7 @@ export const toastControls = defineControls({
 export const selectControls = defineControls({
   multiple: { type: "boolean", defaultValue: false },
   disabled: { type: "boolean", defaultValue: false },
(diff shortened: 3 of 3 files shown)


Hagendorn:
Hi @segunadebayo, hi @cschroeter 👋,

I found a small accessibility issue with the Select component and provided a fix for it in this PR. Specifically, the required state was not being reflected in the accessibility tree, which could cause problems for users relying on assistive technologies.

The fix adds aria-required to the trigger element via getTriggerProps, similar to how aria-invalid is already handled.
Please have a look and let me know if there's anything I should adjust or improve. Thanks!

PR #2790, opened 2025-10-19
feat(angle-slider): implement relative thumb dragging and enhance accessibility checks

anubra266:


segunadebayo:
Thanks for this PR. I've improved it slightly in the main branch.

The key change is to avoid creating a new transition for thumb interactions

PR #2791, opened 2025-10-19
fix(svelte):  optimize merge props

anubra266:


segunadebayo:
Let's go ahead and leave it as is for now. Except there's a verifiable performance benefit, happy to reconsider

PR #2866, opened 2025-11-23
build(lint): cleanup eslint config and packages

StyleShit:
Was reading some code here as I was trying to find out the source of a bug I had with Ark, and noticed that the ESLint config and its dependencies are all over the place, so I opened this small PR to clean it up a little bit

Felt that it'd be kinda overkill to open an issue for this, but I can do it of course if you think otherwise 😄


segunadebayo:
This has been recently done by Claude in the main branch. Thanks again

PR #2870, opened 2025-11-24
feat: add alpine-ts example (WIP)

niLPotential:
## 📝 Description

Adds an example project using Zag with Alpine.js.

`examples/alpine-ts` contains a nitro app compatible with e2e tests.
`lib` contains the adapter logic mostly similar to other frameworks.
`lib/plugin.ts` contains a `usePlugin` hook which connects a Zag machine to Alpine.js as a plugin.
`scripts` contains the client scripts and `server/routes` contains the jsx for SSR.

None

## 🚀 New behavior

```ts
// script.ts
import * as component from "@zag-js/component";
import Alpine from "alpinejs"
import { usePlugin } from "../lib"

// component-name should be kebab-case
Alpine.plugin(usePlugin("component-name", component)
Alpine.start()
```

```html
<!-- index.html -->
<div id="root" x-data x-component-name="{ ...props }">
  <div x-component-name:part-name="{ ...partProps }" />
  <div x-component-name:another-part-name />
  <button x-on:click="$componentName().method()">Click</button>

  <div id="nested" x-component-name.modifier="{ ...props }">
    <div x-component-name:part-name.modifier="{ ...partProps }" />
    <button x-on:click="$componentName('modifier').method()">Click</button>
    <!-- this triggers method of root component from inside nested component -->
    <button x-on:click="$componentName().method()">Click</button>
  </div>
</div>
```

The name passed in when defining the plugin is used through out for directives and magic.
Names and Values of directives should be kebab-case.
The expression, when present, is passed as arguments to the `getComponentPartProps` function.
Other api methods and service are available by calling the corresponding magic. Name of magic is camelCase.
When components need to be more specific (such as when nesting), optional modifiers can be used.
`x-data` must be defined for components to work.

No

## 📝 Additional Information

Last updated: Sep 7, 2026

| e2e | status |
| - | - |
| accordion | ✅ |
| angle-slider | ✅ |
| avatar | ✅ |
| carousel | ✅ |
| cascade-select | ❌ |
| checkbox | ✅ |
| clipboard | ✅ |
| collapsible | ✅ |
| color-picker | ✅ |
| combobox | ✅ |
| context-menu-multiple-trigger | ✅ |
| context-menu | ✅ |
| date-input | ✅ |
| date-picker | ✅ |
| dialog-multiple-trigger | ✅ |
| dialog | ✅ |
| drawer-indent | ✅ |
| drawer-swipe-area | ❌ |
| drawer | ✅ |
| editable | ✅ |
| file-upload | ✅ |
| floating-panel | ✅ |
| hover-card-multiple-trigger | ✅ |
| hover-card | ✅ |
| image-cropper | ✅ |
| listbox | ✅ |
| menu-multiple-trigger | ✅ |
| menu-nested | ✅ |
| menu-option | ✅ |
| menu | ✅ |
| navigation-menu | ✅ |
| number-input | ⚠️ |
| pagination | ✅ |
| password-input | ✅ |
| pin-input | ✅ |
| popover-multiple-trigger | ✅ |
| popover | ✅ |
| radio-group | ✅ |
| rating-group | ✅ |
| select | ⚠️ |
| slider | ✅ |
| splitter | ✅ |
| switch | ✅ |
| tabs | ✅ |
| tags-input | ✅ |
| toast | ✅ |
| toggle-group | ✅ |
| tooltip-multiple-trigger | ✅ |
| tooltip | ✅ |
| tour | ✅ |
| tree-view | ❌ |

- number-input: empty string to valid number transitions fail in e2e, but works with real input or delay between types
- drawer-swipe-area: inconsistent, mostly failing
- select: SelectModel.autofill does not dispatch input event
- cascate-select, tree-view: recusive structure


segunadebayo:
Great work @niLPotential Excited to see this progress

niLPotential:
Some notes on current issues

The main two problems are handling a multitude of items (listbox, select, combobox, etc.) and continuous interactions (dragging in carousel, drawer, floating panel, image cropper, etc.).
I believe both are caused because each part is fragmented.
Currently, each part holds its own state and syncs them to the main service api through effects.
Which means multiple items cause too many effects to run, hurting performance.
Similarly, when state is updated too frequently, race conditions are bound to happen.
Both of which should have been solved through the FSM structure, but again, poor implementation on my end.
 
My best guess to the solution would be to subscribe through some kind of 'computed' function.
But I am not familiar enough with the inner workings of js to have a clear answer right now.
Any advice or suggestions are welcome.

PR #2881, opened 2025-11-30
feat(datepicker): added onVisibleRangeChange callback

IvanKalinin:
## 📝 Description
Added `onVisibleRangeChange ` callback to datepicker api. Useful to determine the current month after clicking next/previous trigger and, for example, load events or different data based on the current calendar page/month.

N/A

## 🚀 New behavior
Triggers `onPageChange` callback on clicking next/previous trigger

No

## 📝 Additional Information
N/A


segunadebayo:
Thanks for contributing this. I was wondering if there was an alternative name to “onPageChange”, given the word “page” doesn’t feel intuitive for a date picker.

Wondering if “onNavigate” or a combo of “onPrevTriggerClick” + “onNextTriggerClick” would work.

IvanKalinin:
> Thanks for contributing this. I was wondering if there was an alternative name to “onPageChange”, given the word “page” doesn’t feel intuitive for a date picker.
> 
> Wondering if “onNavigate” or a combo of “onPrevTriggerClick” + “onNextTriggerClick” would work.

Do you want me to rename it to `onNavigate`?

IvanKalinin:
@segunadebayo Did you make a decision? 
Do you want me to rename the callback to `onNavigate`? Or it is good to go as is, where `onPageChange` makes the perfect sense to me, since the codebase already uses "page" as the entity name. (e.g., getNextPage, getPreviousPage)

https://github.com/chakra-ui/zag/blob/3dbfe4ab5d2eaa9756f2897c92b9b560da9ef9ae/packages/machines/date-picker/src/date-picker.machine.ts#L9-L13

It also makes sense in terms of a physical calendar with pages

segunadebayo:
Yes, you might be right @IvanKalinin 

I was wondering if it's better to attach the `onClick` to the prev/next trigger in your userland code instead of adding a new prop.

The name can be misleading since there might be more scenarios where users might want `onPageChange` to be fired, for example, when the month/year select is used.

IvanKalinin:
@segunadebayo Is there an ETA for when the PR can be merged in Zag and delivered to the Ark UI?
I need the feature to replace a few of my legacy components with the Ark Datepicker.

segunadebayo:
Did you see my message above? @IvanKalinin

IvanKalinin:
@IvanKalinin, I guess I just misunderstood you.
Although the "page" term makes perfect sense to me, I replaced it with onNextTriggerClick and onPrevTriggerClick if you like this more.

IvanKalinin:
@segunadebayo After a little bit more thinking, I decided to implement the feature under the `onVisibleRangeChange` callback.
Should be clear enough.

IvanKalinin:
@segunadebayo, I am wondering if you are ok with the latest implementation under `onVisibleRangeChange`?
There are no on the next/prev trigger click callbacks, so we should introduce a new callback property.

Can this PR be merged?

segunadebayo:
@IvanKalinin I've updated the code to cover more scenarios. It might be a better start; we can improve the use case as we go.

Happy New Year

PR #2937, opened 2026-01-29
feat(menu): support modal behavior

Doesntmeananything:
Addresses: https://github.com/chakra-ui/zag/discussions/1722, https://github.com/chakra-ui/ark/issues/2874

## 📝 Description

Adds support for modal behavior for `menu`.

`Menu` doesn't support modality.

## 🚀 New behavior

The `modal` property can be set to `true` for `menu`, meaning that it will:

- trap focus within its content
- block scrolling on the body
- disable pointer interactions outside the popover
- hide content behind the popover from screen readers

No, the property is set to `false` by default, mirroring the behavior before the change.

## 📝 Additional Information

The change is largely based on how modality is implemented in the `popover`.

### DIFF
--- a/packages/machines/menu/package.json
+++ b/packages/machines/menu/package.json
@@ -27,13 +27,16 @@
   },
   "dependencies": {
     "@zag-js/anatomy": "workspace:*",
+    "@zag-js/aria-hidden": "workspace:*",
     "@zag-js/core": "workspace:*",
-    "@zag-js/dom-query": "workspace:*",
-    "@zag-js/rect-utils": "workspace:*",
-    "@zag-js/utils": "workspace:*",
     "@zag-js/dismissable": "workspace:*",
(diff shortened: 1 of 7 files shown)


segunadebayo:
Thanks for this. Please test that nested menu still works as expected when modal

Doesntmeananything:
Good call. From my tests, if you set just the parent menu to be modal, it works correctly:

https://github.com/user-attachments/assets/963336ef-77ab-45a6-98d6-16d41ea3f348

However, there's some weirdness if you set `modal: true` on submenus - essentially, when moving the pointer over the trigger, the submenu gets constantly opened and closed. Modal submenus don't really seem like a valid use case, but still something to keep in mind:

https://github.com/user-attachments/assets/486598c9-24a2-40ba-8493-8ad5705800b9

AmeerDlshad:
Any update on this please?

PR #3006, opened 2026-03-05
fix(floating-panel): remove `sync: true` from size/position bindables

### LINKED ISSUE #3005: [`@zag-js/floating-panel`]: `sync: true` on size/position bindables causes "Maximum update depth exceeded" when panel content has ResizeObservers
## 🐛 Bug report

Drag-resizing a floating panel intermittently throws `Maximum update depth exceeded` when the panel body contains content with `ResizeObserver` callbacks that trigger React state updates (e.g. chart libraries, auto-sizing components).

## 💥 Steps to reproduce

1. Render a `floating-panel` with content that uses a `ResizeObserver` (e.g. a chart component, or a custom hook like the one below)
2. Drag-resize the panel rapidly

```tsx
// Inside floating panel content
function AutoSizingContent() {
  const ref = useRef<HTMLDivElement>(null);
  const [size, setSize] = useState({ width: 0, height: 0 });

  useEffect(() => {
    const el = ref.current;
    if (!el) return;
    const observer = new ResizeObserver(([entry]) => {
      const { width, height } = entry.contentRect;
      setSize({ width, height });
    });
    observer.observe(el);
    return () => observer.disconnect();
  }, []);

  return <div ref={ref}>Content: {size.width}x{size.height}</div>;
}
```

## 🧐 Expected behavior

Drag-resize should work without errors regardless of what content is rendered inside the panel body.

## 🔍 Root cause analysis

The `size` and `position` bindables in the floating-panel machine are declared with `sync: true`:

```js
// floating-panel.machine.ts
size: bindable(() => ({
  // ...
  sync: true,
})),
position: bindable(() => ({
  // ...
  sync: true,
})),
```

In the React adapter, `sync: true` wraps `setState` in `flushSync`:

```js
// @zag-js/react — bindable.ts
set(value2) {
  const exec = props().sync ? flushSync : identity;
  exec(() => setFn(value2));
},
```

During resize drag, each pointer-move event calls `setSize`, which calls `context.set("size")` and `context.set("position")` — both triggering `flushSync`. This forces React to synchronously commit (render + DOM update + effects). If the panel's content has `ResizeObserver` callbacks that call `setState`, those fire within the same synchronous frame, causing a cascading render loop that hits React's max update depth.

**Why it's intermittent:** It depends on whether `ResizeObserver` callbacks fire within the same synchronous execution frame as the `flushSync` commit, which varies with browser layout timing and drag speed.

## 🧭 Possible solution

Remove `sync: true` from the `size` and `position` bindables. This is safe because:

1. **`setSizeStyle` already uses `queueMicrotask`**, so the CSS custom property update for width/height is already asynchronous — `flushSync` doesn't make it any faster
2. **React 18+ automatically batches** state updates within event handlers and commits at the end of the handler, still within the same animation frame
3. **`setSize` computes from `prevSize`/`prevPosition`** (snapshots taken at drag start), not from `context.get("size")`, so there is no stale-value drift from batched updates

The visual difference is imperceptible — both paths result in the CSS custom properties being updated and painted in the same frame.

If `sync: true` is needed for other frameworks (Solid, Vue, Svelte), consider making it configurable per-framework in the React adapter, or adding a re-entrancy guard in the `setSize`/`setPosition` actions.

## 🌍 System information

| Software         | Version(s)  |
| ---------------- | ----------- |
| Zag Version      | 1.35.3      |
| Framework        | React 19    |
| Browser          | Chrome, Safari |
| Operating System | macOS       |

## 📝 Additional information

Workaround: patching `@zag-js/floating-panel` to remove `sync: true` from both bindables resolves the issue with no observable visual regression during drag/resize.

cminn10:
## Summary

Remove `sync: true` from the `size` and `position` bindables in the floating-panel machine to fix intermittent "Maximum update depth exceeded" errors in React.

Fixes #3005

## Problem

`sync: true` causes the React adapter's `useBindable` to wrap `setState` in `flushSync`. During drag-resize, the `setSize` action calls `context.set("size")` and `context.set("position")` — both triggering `flushSync`. This forces React to synchronously commit renders. If the panel's content contains `ResizeObserver` callbacks that trigger state updates (common with chart libraries, auto-sizing components, etc.), those fire within the same synchronous frame, causing a cascading render loop.

## Why removing `sync: true` is safe

1. **`setSizeStyle` already uses `queueMicrotask`** — the CSS custom property update for width/height is already asynchronous, so `flushSync` doesn't actually make the visual update any faster
2. **React 18+ automatically batches** state updates within event handlers and commits at the end of the handler, still within the same animation frame
3. **`setSize` computes from `prevSize`/`prevPosition`** (snapshots taken at drag start), not from `context.get("size")`, so there is no stale-value drift from batched updates
4. **React 17 / legacy mode** — native event handlers (which `trackPointerMove` uses via `document.addEventListener`) are already not batched, so `flushSync` is redundant there too

The visual difference is imperceptible — both paths result in the CSS custom properties being updated and painted in the same frame.

## Change

Two lines removed from `packages/machines/floating-panel/src/floating-panel.machine.ts`:

```diff
 size: bindable<Size>(() => ({
   defaultValue: prop("defaultSize"),
   value: prop("size"),
   isEqual: isSizeEqual,
-  sync: true,
   hash(v) { ... },
   onChange(value) { ... },
 })),
 position: bindable<Point>(() => ({
   defaultValue: prop("defaultPosition"),
   value: prop("position"),
   isEqual: isPointEqual,
-  sync: true,
   hash(v) { ... },
   onChange(value) { ... },
 })),
```

### DIFF
--- a/packages/machines/floating-panel/src/floating-panel.machine.ts
+++ b/packages/machines/floating-panel/src/floating-panel.machine.ts
@@ -59,7 +59,6 @@ export const machine = createMachine<FloatingPanelSchema>({
         defaultValue: prop("defaultSize"),
         value: prop("size"),
         isEqual: isSizeEqual,
-        sync: true,
         hash(v) {
           return `W:${v.width} H:${v.height}`
         },
@@ -71,7 +70,6 @@ export const machine = createMachine<FloatingPanelSchema>({
         defaultValue: prop("defaultPosition"),
         value: prop("position"),
         isEqual: isPointEqual,
-        sync: true,
         hash(v) {
           return `X:${v.x} Y:${v.y}`
         },


segunadebayo:
Thanks for this PR. The fix was beyond just removing `sync: true` and I've addressed it

PR #3014, opened 2026-03-07
fix(pin-input): focus first empty slot on empty-field click

### LINKED ISSUE #2999: Clicking any PinInput.Input should focus the first Input when empty
### Description

When I click on any box of an empty PinInput group, I expect it to extend keyboard focus to the _first_ element in the set, not the one clicked.

This is an accessibility issue for low mobility users, who would expect to be able to tap any of the boxes and type in all items, not just starting from the one tapped.

Generally, I would assume focus would always move to the _first_ empty box when tapping _any_ empty box.

Scenario: all empty:
```
[ ] [ ] [ ] [ ]
         ^ tap here
# Should result in
[ ] [ ] [ ] [ ]
 ^ focus is placed here
```

Scenario: not all empty:
```
[1] [2] [ ] [ ]
             ^ tap here
# Should result in
[1] [2] [ ] [ ]
         ^ focus is placed here
```

Scenario: not all empty:
```
[1] [2] [ ] [ ]
     ^ tap here
# Should result in
[1] [2] [ ] [ ]
     ^ focus is placed here
```

https://codesandbox.io/p/devbox/black-hill-glp8wv

### Steps to Reproduce

1. Go to https://ark-ui.com/docs/components/pin-input
2. In the first example, click the 3rd box from the left
3. Start typing. Notice that when you type more than 2 items, it won't fill back to the first two boxes.

### Ark UI Version

5.34.0

### Framework

- [x] React
- [x] Solid
- [x] Vue
- [x] Svelte

### Browser

all

### Additional Information

_No response_

SergioChan:
Closes #2999

## 📝 Description

- update `setFocusedIndex` to route focus to the first empty pin slot when the clicked input is empty
- keep existing behavior for clicks on already-filled inputs

- clicking an empty slot focuses that exact slot, even when earlier slots are still empty
- this can make keyboard entry start from the middle and leave earlier cells unfilled

## 🚀 New behavior

- clicking any empty slot now moves focus to the first empty slot in the group
- clicking a filled slot still focuses the clicked slot

No

## 📝 Additional Information

Validated locally with:
- `pnpm --filter @zag-js/pin-input lint`
- `pnpm --filter @zag-js/pin-input typecheck`

### DIFF
--- a/packages/machines/pin-input/src/pin-input.machine.ts
+++ b/packages/machines/pin-input/src/pin-input.machine.ts
@@ -189,8 +189,19 @@ export const machine = createMachine({
       clearFocusedIndex({ context }) {
         context.set("focusedIndex", -1)
       },
-      setFocusedIndex({ context, event }) {
-        context.set("focusedIndex", event.index)
+      setFocusedIndex({ context, event, computed }) {
+        const value = computed("_value")
+        const clickedIndex = event.index
+
+        // If user taps an empty slot, move focus to the first empty slot.
+        // This keeps keyboard entry starting from the earliest missing value.
+        if (value[clickedIndex] === "") {
+          const firstEmptyIndex = value.findIndex((item) => item === "")
+          context.set("focusedIndex", firstEmptyIndex === -1 ? clickedIndex : firstEmptyIndex)
+          return
+        }
+
+        context.set("focusedIndex", clickedIndex)
       },
       setValue({ context, event }) {
         const value = fill(event.value, context.get("count"))


segunadebayo:
Thanks for this PR. I’ve tested it and observed that it broken some other interactions. 

I don’t think hijacking focus ends well most of the time. I’d recommend we keep it as is for now.

We’ll reconsider this in the future.

paularmstrong:
> Thanks for this PR. I’ve tested it and observed that it broken some other interactions.
> 
> I don’t think hijacking focus ends well most of the time. I’d recommend we keep it as is for now.
> 
> We’ll reconsider this in the future.

Can you elaborate what "other interactions" this broke? Maybe that is fixable and the PR could be updated.

anubra266:
@paularmstrong sample.

https://github.com/user-attachments/assets/dc312026-4a87-45c1-9b88-3a021fa5119e

But that said, we don't wanna hijack focus, you can implement that for now i userland, with `onFocus` or `onClick` of your inputs.

paularmstrong:
> @paularmstrong sample.
> CleanShot.2026-03-19.at.10.01.25.mp4

It looks like this is not implemented properly per the original intent.

> we don't wanna hijack focus

I disagree — I would expect the focus moving to the first empty box. An OTP input should work like a single input, just with boxes highlighting individual items.

For example, if it were instead implemented as a single input with masking:

```
[******]
   ^ click third item
[******]
 ^ first item is where cursor goes
```

The focus is irrelevant and should move to the first empty item. Splitting them into separate boxes is a purely _visual_ treatment:

```
[*][*][*]][*][*][*]
       ^ click third item
[*][*][*]][*][*][*]
 ^ first item is where cursor goes — and so focus would follow
```

segunadebayo:
Thanks for this PR. I've gone ahead to improve the pin input's UX even more. It also includes this fix as well.

PR #3062, opened 2026-04-06
fix(navigation-menu): drop default aria-label

jramke:
## 📝 Description

Removes the default `aria-label` from the `navigation-menu` machine. As it caused redundant announcement and was a little agressive.

- `aria-label="Main Navigation"` does not really work on a plain `div`. Role `navigation` is not forced.
- Normally users should use a `nav` or `role="navigation"` themself and this would result in duplicate announcement of the word "Navigation" because it is triggered by the role and the label
- The `aria-label` is not needed when theres only one navigation on the page and users with a non english site would then need to override the label because of this

## 🚀 New behavior

- Doesnt set a default label at all
- Matches consistency and expectaions across the machine since we also dont force the `nav`/`role="navigation"` or `ul`/`li` semantics

No
- users with a single instance will get the same semantic/a11y behavior
- users with more instances would previously got a broken a11y behavior because of duplicate nav labels

## 📝 Additional Information
Maybe its worth adding a note in the docs for the semantic elements for some parts:
- root -> `<nav>`
- list -> `<ul>`
- item -> `<li>`

And the indicator element should be moved out of the ul if possible i guess since it would result in invalid html

### DIFF
--- a/packages/machines/navigation-menu/src/navigation-menu.connect.ts
+++ b/packages/machines/navigation-menu/src/navigation-menu.connect.ts
@@ -66,7 +66,7 @@ export function connect<T extends PropTypes>(
       return normalize.element({
         ...parts.root.attrs,
         id: dom.getRootId(scope),
-        "aria-label": translations.rootLabel,
+        "aria-label": translations?.rootLabel ? translations.rootLabel : undefined,
         "data-orientation": prop("orientation"),
         dir: prop("dir"),
         style: {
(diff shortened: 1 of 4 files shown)


packages/machines/navigation-menu/src/navigation-menu.connect.ts
```diff
@@ -66,7 +66,7 @@ export function connect<T extends PropTypes>(
       return normalize.element({
         ...parts.root.attrs,
         id: dom.getRootId(scope),
-        "aria-label": translations.rootLabel,
+        "aria-label": translations?.rootLabel ? translations.rootLabel : undefined,
```
segunadebayo:
```suggestion
        "aria-label": translations?.rootLabel,
```

PR #3066, opened 2026-04-08
test: extend e2e tests for date input granularity

bjoluc:
## 📝 Description

According to the TS types, the date input (which is an awesome addition by the way!) supports `"month"` and `"year"` granularities, which are not available in the shared controls yet and do not seem implemented in general. If those are not planned, feel free to close this and update the types. Otherwise, this PR adds
- (failing) tests for the `"month"` and `"year"` granularities
- assertions about segments that should not be present in existing granularity tests

`"month"` and `"year"` granularities do not work in date input, are not included in shared controls, and do not have tests. 

## 🚀 New behavior

Shared controls are updated and (failing) tests are there.

No

### DIFF
--- a/e2e/date-input.e2e.ts
+++ b/e2e/date-input.e2e.ts
@@ -223,9 +223,52 @@ test.describe("date-input [single]", () => {
     await I.seeSegmentFocused("month")
   })
 
+  test("[input] year granularity", async () => {
+    await I.clickControls()
+    await I.controls.select("granularity", "year")
+
+    // Month (and finer) segments should not be present
+    await I.seeSegmentNotPresent("month")
+    await I.seeSegmentNotPresent("day")
+    await I.seeSegmentNotPresent("minute")
+    await I.seeSegmentNotPresent("seconds")
+
+    // Year segment should be present and focusable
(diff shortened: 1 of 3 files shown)


segunadebayo:
Good catch. I've updated the types to not include month/year in granularity.

A segmented date input with only 1–2 editable fields (just year, or month + year) is over-engineered. If someone needs month-only or year-only selection, a <select>, a number input, or a calendar picker with a month/year view is a better fit.

bjoluc:
Yup, agree. Thanks!

e2e/date-input.e2e.ts
```diff
@@ -223,10 +223,53 @@ test.describe("date-input [single]", () => {
...
+  })
+
   test("[input] hour granularity", async () => {
     await I.clickControls()
     await I.controls.select("granularity", "hour")
 
+    // Minute and second segments should not be present
+    await I.seeSegmentNotPresent("minute")
+    await I.seeSegmentNotPresent("seconds")
+
```
bjoluc, line 272: If you have a minute, stuff like this wouldn't hurt anyway 🙂

PR #3069, opened 2026-04-09
fix(date-utils): fix CalendarDateTime handling in isDateEqual

bjoluc:
## 📝 Description

`isDateEqual` from `@zag-js/date-utils` accepts two `DateValue`s, i.e. `CalendarDate | CalendarDateTime | ZonedDateTime` from `@internationalized/date`. The time components of `CalendarDateTime` and `ZonedDateTime` are currently ignored in the comparison. This PR fixes that.

`isDateEqual` from `@zag-js/date-utils` ignores the time components of `CalendarDateTime` and `ZonedDateTime` values. This prevents the `onValueChange` callback to be invoked on time changes in the date input machine:

https://github.com/chakra-ui/zag/blob/b5fe9903f3168746b2c5d5dd2bd38edad0c51043/packages/machines/date-input/src/date-input.machine.ts#L115

The current implementation of `isDateEqual` uses `isSameDay` for the comparison.

## 🚀 New behavior

The comparison is performed by `toCalendarDateTime(dateA).compare(dateB) === 0`. While this does consider the time portion of `CalendarDateTime` and `ZonedDateTime` values, it ignores the time zone of `ZonedDateTime`s. I couldn't find a use case for this in the code base yet, but let me know if this is a concern. Cheers!

No

## 📝 Additional Information

At a first glance, the tests at `packages/utilities/date-utils/tests/` do not seem to be included in any existing pnpm script and/or workflow. Am I wrong?

### DIFF
--- a/packages/utilities/date-utils/src/assertion.ts
+++ b/packages/utilities/date-utils/src/assertion.ts
@@ -1,9 +1,9 @@
-import { isSameDay } from "@internationalized/date"
+import { isSameDay, toCalendarDateTime } from "@internationalized/date"
 import type { DateAvailableFn, DateValue } from "./types"
 
 export function isDateEqual(dateA: DateValue | null | undefined, dateB?: DateValue | null) {
   if (dateA == null || dateB == null) return dateA === dateB
-  return isSameDay(dateA, dateB)
+  return toCalendarDateTime(dateA).compare(dateB) === 0
 }
 
 export function isDateUnavailable(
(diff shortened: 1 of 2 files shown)


segunadebayo:
Thanks for this PR. I pushed a better fix for this so it doesn't break other components

bjoluc:
Thanks for fixing this! Out of curiosity: I fail to see how
```ts
if (!("hour" in dateA) && !("hour" in dateB)) return isSameDay(dateA, dateB)
return toCalendarDateTime(dateA).compare(toCalendarDateTime(dateB)) === 0
```
would behave different than my `return toCalendarDateTime(dateA).compare(dateB) === 0`?
Anyway, now that I've already taken the time to add unit tests, would you mind pulling these in too?
Thanks again!

PR #3092, opened 2026-04-20
fix(popover): restart tracked positioning for lazy-mounted positioners

aleclarson:
## Description
Restarts the tracked popover positioning session when a lazy-mounted positioner appears, so observers can attach after the floating node exists.

Controlled, lazy-mounted popovers can miss the initial tracked positioning pass because the positioner is not in the DOM yet.

## New behavior
The machine now has a dedicated `restartPositioning()` path for rebuilding the tracked positioning session after mount. `reposition()` stays as the one-shot manual path.

## Is this a breaking change (Yes/No):
No

## Additional Information
This is still a draft. I did not add tests, and I won't have time to push this through to the finish from here. If this direction looks useful, it'd be great if someone could pick it up. Thanks for your time.

### DIFF
--- a/packages/machines/popover/src/popover.connect.ts
+++ b/packages/machines/popover/src/popover.connect.ts
@@ -35,6 +35,9 @@ export function connect<T extends PropTypes>(service: PopoverService, normalize:
     reposition(options = {}) {
       send({ type: "POSITIONING.SET", options })
     },
+    restartPositioning() {
+      send({ type: "POSITIONING.RESTART" })
+    },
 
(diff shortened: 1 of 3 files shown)


segunadebayo:
Hey! Could you show a reproduction of the issue you’re attempting to fix?

aleclarson:
It seems to be an issue with Preact only. Though I wasn't using `@zag-js/preact` in place of `@zag-js/react`. Let me try that and I'll see if this is still an issue or not. ✌️

PR #3102, opened 2026-04-29
feat(infinite-scroll): add IntersectionObserver utility

Adebesin-Cell:
## 📝 Description

New `@zag-js/infinite-scroll` package. A framework-agnostic utility that watches a sentinel element and calls `onLoadMore` when it scrolls into view.

It exports `createInfiniteScroll(options)` (returns a cleanup) and `getSentinelProps()`.

Infinite scroll is stateless from the machine's side: observe a sentinel, call a function. The loading, hasMore, and error state already lives in the consumer or in `@zag-js/async-list`, so a machine would be overkill. But the `IntersectionObserver` wiring is fiddly enough to be worth sharing.

Use it to:

- Paginate data with `@zag-js/async-list`
- Load large lists with `@zag-js/virtualizer` (sentinel is the last virtual row)
- Load older messages in a chat (sentinel goes at the top)

No shared infinite-scroll utility. Each consumer wires up its own `IntersectionObserver`.

## 🚀 New behavior

`createInfiniteScroll` observes the sentinel and fires `onLoadMore` when it enters the scroll root. Details it handles for you:

- `scrollerEl` defaults to the sentinel's nearest scrollable ancestor, falling back to the viewport.
- `rootMargin` applies to all four sides, so a top sentinel (chat) and horizontal scrollers work with no extra options.
- `threshold` is pixels as a `number`, or any CSS margin token as a `string` (so `"100%"` gives a ratio-style threshold).
- `getSentinelProps()` bundles the non-obvious bits: a 1×1 size (zero-size sentinels don't fire reliably), `inert`, `aria-hidden`, and `pointer-events: none`.

Consumers re-call `createInfiniteScroll` when `hasMore`, `loading`, or items change. This is deliberate: with a large `rootMargin`, content changes inside the margin window don't re-fire the callback on their own, since visibility hasn't changed. The examples show the `useEffect` deps for this.

Examples added: basic, with `async-list`, with `virtualizer`, under a new `infinite-scroll` route.

Left out on purpose:

- No data fetching, error UI, or pull-to-refresh. Those are consumer concerns.
- No `viewportRatio: number` shorthand and no axis-specific margins. Both can be added later without breaking the API.

No. New package.

## 📝 Additional Information

- `pnpm typecheck` and `pnpm lint` pass for the new package.
- `pnpm build` produces ESM, CJS, and `.d.ts`.
- I read through the three example pages but haven't booted the dev server. A sanity check there would help.


segunadebayo:
Closing. I'll replace this with a full state machine shortly.

PR #3125, opened 2026-05-18
fix(combobox): handle Enter correctly when item highlighted

ylc395:
## 📝 Description
See below

1. use a opening Combobox with `allowCustomValue` with a collection
2. type something
3. select an item with Enter key
4. **default behavior triggered** (but if you select an item by clicking, default behavior won't be triggered)

## 🚀 New behavior

1-3: the same
4. **default behavior won't be triggered**

No


segunadebayo:
Thanks for this PR. I've pushed a slightly better fix in main branch

PR #3128, opened 2026-05-19
fix(number-input): avoid clamping empty value on blur

MoonBrillante:
Related to chakra-ui/chakra-ui#10820

## 📝 Description

This fixes an inconsistent blur behavior in `number-input` when the input is cleared and `min` is greater than `0`.

The original issue was reported in Chakra UI, but the underlying behavior comes from `@zag-js/number-input`.

## 🚛 Current behavior

When the input is cleared, the value becomes empty, but the blur logic can still enter the clamp path because it only checks:

```ts
and("clampValueOnBlur", not("isInRange"))

### DIFF
--- a/packages/machines/number-input/src/number-input.machine.ts
+++ b/packages/machines/number-input/src/number-input.machine.ts
@@ -181,7 +181,7 @@ export const machine = createMachine({
         },
         "INPUT.BLUR": [
           {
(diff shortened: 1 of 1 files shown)


segunadebayo:
Thanks for this fix. I've improved it and added tests in the main branch.

PR #3157, opened 2026-05-30
fix: correct pin input filled state

### LINKED ISSUE #3156: PinInput: data-filled prop is always present on first render
# 🐛 Bug report

I did notice `pin-input`, when first-rendered always has `data-filled` attribute set. It will behave as normal after changing the value by any means.

This happens whenever the value is controlled or not.

## 💥 Steps to reproduce

1. Go to the [Ark UI Pin Input example](https://ark-ui.com/docs/components/pin-input) and reach the first example;
2. Inspect the first input;
3. First Result: Every input has `data-filled` set;
4. Edit the value;
6. Second Result: `data-filled` now behaves as expected.

## 💻 Link to reproduction

CodeSandbox reproduction: —

## 🧐 Expected behavior

- `data-filled` should only be set where that input part is filled, no matter if it is controlled or not.

## 🧭 Possible Solution

Looking at the code, The logic is well wired, however I believe the bug is related to the `value` prop being `undefined` instead of `""` while checking for the specific index [(Here's the part)](https://github.com/chakra-ui/zag/blob/main/packages/machines/pin-input/src/pin-input.connect.ts#L126).

## 🌍 System information

| Software                 | Version(s)      |
| ------------------------ | --------------- |
| Zag Version              | 1.41.1          |
| Ark UI (React) Version   | 5.37.0          |
| Browser                  | Chrome          |
| Operating System         | macOS / Windows |

## 📝 Additional information

As a workaround, I'm manually adding a `data-is-filled` prop to the component and working with it instead.

```tsx
type AppPinInputProps = PinInput.InputProps

const AppPinInput: React.FC<AppPinInputProps> = ({ index, ...delegated }) => {
  const { valueAsString } = usePinInputContext()

  const isFilled = !!valueAsString.at(index) // or Boolean(valueAsString.at(index))

  return (
    <PinInput.Input
      {...delegated}
      data-is-filled={isFilled || undefined}
      index={index}
    />
  )
}
```

segunadebayo:
Closes #3156

## 📝 Description

Fix pin input slots incorrectly receiving `data-filled` on the first render when their value has not been initialized yet.

Empty pin input slots could read from the raw value array before it was normalized, so `undefined !== ""` marked every slot as filled on initial render.

## 🚀 New behavior

Input props use the normalized computed value for both `data-filled` and the initial input value, so only slots with actual values receive `data-filled`.

No

## 📝 Additional Information

Added Playwright coverage for uncontrolled and controlled initial empty states.

## ✅ Tests

- `pnpm build`
- `pnpm e2e-react e2e/pin-input.e2e.ts`

<div><a href="https://cursor.com/agents/bc-176e601c-3b17-4148-bed2-98fe0cef8500"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-176e601c-3b17-4148-bed2-98fe0cef8500"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/.changeset/gentle-pins-float.md
+++ b/.changeset/gentle-pins-float.md
@@ -1,0 +1,5 @@
+---
+"@zag-js/pin-input": patch
+---
+
+Fix initial `data-filled` state for empty pin input fields.
--- a/e2e/pin-input.e2e.ts
+++ b/e2e/pin-input.e2e.ts
@@ -9,6 +9,12 @@ test.describe("pin input", () => {
     await I.goto()
   })
 
+  test("data-filled: should not be set on empty inputs on first render", async () => {
+    await I.dontSeeInputHasAttribute(1, "data-filled")
+    await I.dontSeeInputHasAttribute(2, "data-filled")
+    await I.dontSeeInputHasAttribute(3, "data-filled")
+  })
+
   test("on type: should move focus to the next input", async () => {
     await I.fillInput(1, "1")
     await I.seeInputIsFocused(2)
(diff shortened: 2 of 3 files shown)


yuriuliam (commented):
Looks good to me!

PR #3159, opened 2026-05-31
fix(signature-pad): forward dir prop to visible parts

### LINKED ISSUE #3158: signature-pad: `dir` prop is exposed but never applied in connect
## Description

The Signature Pad machine accepts a `dir` prop, but `connect` never forwards it to the DOM. Passing `dir="rtl"` has no effect on the rendered output.

- [`signature-pad.props.ts`](https://github.com/chakra-ui/zag/blob/main/packages/machines/signature-pad/src/signature-pad.props.ts) includes `"dir"`.

### Connect — missing on all visible parts

In [`signature-pad.connect.ts`](https://github.com/chakra-ui/zag/blob/main/packages/machines/signature-pad/src/signature-pad.connect.ts), none of the visible connect functions set `dir: prop("dir")`:

| Connect part | Anatomy part |
|---|---|
| `getRootProps` | `root` |
| `getLabelProps` | `label` |
| `getControlProps` | `control` |
| `getSegmentProps` | `segment` |
| `getSegmentPathProps` | `segmentPath` |
| `getGuideProps` | `guide` |
| `getClearTriggerProps` | `clearTrigger` |

`prop("dir")` is not referenced anywhere in connect or the machine logic.

`getHiddenInputProps` is intentionally excluded — other machines (`checkbox`, `switch`, `pin-input`, etc.) also omit `dir` on hidden inputs.

### Expected behavior

Match other machines : add `dir: prop("dir")` to each visible connect part.

### Suggested fix

Add `dir: prop("dir")` to all 7 visible connect parts in `signature-pad.connect.ts`.

MoonBrillante:
Closes #3158

## Description

This PR forwards the `dir` prop to the visible Signature Pad parts so RTL direction can be applied correctly.

## Changes

- Added `dir: prop("dir")` to visible Signature Pad parts
- Used spread syntax for SVG/path parts to keep the types valid

## Breaking change

No

## Notes

I could not run the local typecheck because the current pnpm version requires Node.js >= 22.13, while my local Node.js version is 20.19.5.

### DIFF
--- a/.changeset/signature-pad-dir.md
+++ b/.changeset/signature-pad-dir.md
@@ -1,0 +1,5 @@
+---
+"@zag-js/signature-pad": patch
+---
+
+Forward the `dir` prop to visible signature pad parts.
--- a/packages/machines/signature-pad/src/signature-pad.connect.ts
+++ b/packages/machines/signature-pad/src/signature-pad.connect.ts
@@ -35,6 +35,7 @@ export function connect<T extends PropTypes>(
     getLabelProps() {
       return normalize.label({
         ...parts.label.attrs,
+        dir: prop("dir"),
         id: dom.getLabelId(scope),
         "data-disabled": dataAttr(disabled),
         "data-required": dataAttr(required),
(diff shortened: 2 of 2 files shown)


segunadebayo:
Thanks for this, PR. I've added it to the main branch

PR #3163, opened 2026-06-03
Adds initialFocusSkip option to focus-trap

### LINKED ISSUE #3161: Support data-no-autofocus in focusTrap or similar
# 🚀 Feature request

It would be great if there was a property that could be added to elements (I'm suggesting `data-no-autfocus`) that would cause focusTrap to skip over those elements when setting initialFocus.

When opening dialog's/popovers/etc there are some elements that you may never want to be auto-focused, even if they are the first focusable element, tooltips probably being the common use-case.

While its possible at the moment to manually set initialFocus to some other element, this can get repetitive and it would be a lot nicer if we could also just mark some elements as not being auto-focusable and have focusTrap skip over them when looking for the initial focus.

alex-e-leon:
Closes https://github.com/chakra-ui/zag/issues/3161

## 📝 Description

Adds a new option to focus-trap that allows one to define elements that should avoid being the initial focused element when a focus trap initializes. I've added it as a new option so that the feature is opt-in and the regular behaviour should remain unchanged. I'm also open to other approaches but thought a proof of concept would help.

the initial focused element in focus-trap must be manually specified.

## 🚀 New behavior

elements can now be marked to avoid being the initial focused elements - allowing elements like tooltips or other elements with distracting focused states to mark themselves as preferring not to be initially focused.

No

### DIFF
--- a/packages/utilities/focus-trap/src/focus-trap.ts
+++ b/packages/utilities/focus-trap/src/focus-trap.ts
@@ -503,6 +503,28 @@ export class FocusTrap {
     }
   }
 
+  private getInitialTabbableNode = () => {
+    const firstGroup = this.state.tabbableGroups[0]
+    const skipSelector = this.config.initialFocusSkip
+    if (!firstGroup) return undefined
+    if (!skipSelector) return firstGroup.firstTabbableNode
+
+    for (const group of this.state.tabbableGroups) {
+      for (const node of group.tabbableNodes) {
+        let matches: boolean
+        try {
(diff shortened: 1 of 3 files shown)


segunadebayo:
Just pushed support for this.

In the next release, this lands as `data-no-autofocus` / `data-autofocus` on `getInitialFocus`, with dialog/drawer wired through that helper instead of `initialFocusSkip` on focus-trap.

Thanks for the PR. Closing based on that approach.

Issue #3173, opened 2026-06-10
[Toast] Height measurement issue when creating multiple toasts sync


Hwacc:
# 🐛 Bug report

## Description

With `overlap: true`, sibling toasts get a `scale` transform on the root element. `measureHeight` and `trackHeight` measure height via `getBoundingClientRect().height` on that same element, so the stored value includes the scale. It is saved to `heights` and used as `--initial-height`. After the frontmost toast is dismissed and the stack expands on hover, the remaining toast renders at the wrong height.

If the root has CSS `transition` on `scale` / `height`, measurement without disabling transitions may also read interpolated values.

**Affected:** `toast.machine.ts` — `measureHeight`, `trackHeight` (`syncHeight`).

## 💻 Link to reproduction

See: [toast demo](https://stackblitz.com/edit/dvees97v?file=src%2FToast.tsx)

## 💥 Steps to reproduce

1. Click `Show Double Toast` Button
2. Close `#2 Toast` on the top
3. Then hover on `#1 Toast`,  the toast body height is flickering

## 🧐 Expected behavior

Measured height should be the untransformed layout height, independent of overlap `scale` at measure time.

## 🧭 Possible Solution

In both `measureHeight` and `trackHeight`, temporarily set `transition: 'none'`, neutralize `scale` / `translate` / `transform`, set `height: 'auto'`, then read `getBoundingClientRect().height` and restore inline styles.

## 🌍 System information

| Software    | Version(s)   |
| ----------- | ------------ |
| Zag Version | `@zag-js/toast@1.41.1` |
| Browser     | Chromium     |
| OS          | Windows 10   |

## 📝 Additional information

Root cause: overlap `scale` is applied to the same element that height measurement reads from (`dom.getRootEl`).

segunadebayo:
Fixed in #3178 — toast height is now measured from the untransformed element, so the overlap `scale` transform no longer leaks into the stored height. Verified in the browser across React, Preact, Svelte, Vue, and Solid.

PR #3174, opened 2026-06-10
Optimize height measurement using measureUntransformedHeight

Hwacc:
This PR is aimed to address issue #3173


segunadebayo:
Thanks for the contribution, @Hwacc! 🙏 This is addressed in #3178, which takes the same approach (neutralizing the transform before measuring) and has been verified across React, Preact, Svelte, Vue, and Solid. Closing as superseded — really appreciate you digging into the root cause.

PR #3177, opened 2026-06-12
fix(dismissable): restore layer style mirror after framework style rewrites

### LINKED ISSUE #3148: `Drawer.Backdrop` loses layer stack variables after dragging content
### Description

When i drag content, i expect that the variables don't disappear, but currently they disappear and `z-index: calc(100 + var(--layer-index))` calculation breaks.

https://github.com/user-attachments/assets/073c361b-794d-4a09-bfbd-0cf47fe9a163

https://stackblitz.com/edit/ivbcxkqr?file=src%2FApp.svelte&showSidebar=0

### Steps to Reproduce

1. Click open.
2. Drag content.
3. Open backdrop in the devtools.
4. Notice there are no `--layer-index`, `--nested-layer-count` and `--z-index` variables.

### Ark UI Version

5.22.0

### Framework

- [ ] React
- [ ] Solid
- [ ] Vue
- [x] Svelte

### Browser

_No response_

### Additional Information

_No response_

spokodev:
Fixes #3148

### Problem

The layer stack mirrors its metadata (`--layer-index`, `--nested-layer-count`, `--z-index`, `data-nested`, `data-has-nested`) onto style targets like `Drawer.Backdrop` imperatively via `style.setProperty`, but only when layers are added or removed. Frameworks own the same elements' `style` and replace the inline style wholesale on re-render — for the drawer, every drag pointermove updates `--drawer-swipe-progress` on the backdrop, so the first drag frame erases the mirrored variables and `z-index: calc(100 + var(--layer-index))` breaks for the rest of the layer's lifetime.

### Fix

Watch each mirrored element's `style` attribute with a `MutationObserver` and re-apply the metadata when it disappears while the layer is still in the stack. The observer is a no-op when the metadata is present (no loops: re-applying triggers one callback that sees `--layer-index` set and returns), keeps whatever the framework wrote (`--drawer-swipe-progress` survives), and is disconnected when the layer is removed so cleared styles stay cleared.

### Tests

`restores mirrored metadata after a framework rewrite of the style attribute` fails on `main` (`--layer-index` stays empty after `setAttribute("style", ...)`) and passes with this change. A second test pins that removal still clears the mirror and the observer stops restoring. Dismissable suite: 15 passed.

### DIFF
--- a/.changeset/fix-layer-style-mirror-rewrite.md
+++ b/.changeset/fix-layer-style-mirror-rewrite.md
@@ -1,0 +1,8 @@
+---
+"@zag-js/dismissable": patch
+---
+
+Fix issue where layer stack metadata (`--layer-index`, `--nested-layer-count`,
+`--z-index`) disappeared from style targets like `Drawer.Backdrop` when a
+framework re-render replaced the inline style, e.g. while dragging drawer
+content.
--- a/packages/utilities/dismissable/src/layer-stack.ts
+++ b/packages/utilities/dismissable/src/layer-stack.ts
@@ -106,9 +106,11 @@ export const layerStack = {
     if (index < 0) return
 
     const layer = this.layers[index]
+    unobserveLayerStyleMirror(node)
     layer.styleTargets?.forEach((getTarget) => {
       const target = getTarget()
       if (target) {
+        unobserveLayerStyleMirror(target)
         clearLayerStyleMirror(target)
       }
     })
@@ -139,12 +141,14 @@ export const layerStack = {
   syncLayers() {
     this.layers.forEach((layer, index) => {
       applyLayerStackMetadata(layer, index, layer.node)
+      observeLayerStyleMirror(layer, layer.node)
       layer.styleTargets?.forEach((getTarget) => {
         const target = getTarget()
         if (!target || target === layer.node) return
         applyLayerStackMetadata(layer, index, target)
         const { zIndex } = getComputedStyle(layer.node)
         target.style.setProperty("--z-index", zIndex)
+        observeLayerStyleMirror(layer, target)
       })
     })
   },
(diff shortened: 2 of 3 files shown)


kalisaNkevin:
Went through this one in detail — the fix and the new test coverage both hold up.

The approach (a `MutationObserver` per style target, keyed by a `WeakMap`) correctly solves #3148: it watches the `style` attribute and reapplies the layer-stack CSS vars when a framework re-render wipes them, which is exactly what was happening to `Drawer.Backdrop` during drag. The self-healing guard (`if target.style.getPropertyValue("--layer-index") !== "") return`) is what keeps this from looping on its own writes, so that edge case looks handled deliberately rather than by luck.

The test file is also a real addition on its own merits — `layer-stack.ts` had zero unit tests before this, and the new suite covers `add`/`remove`/`dismiss`, pointer-blocking, branches, the `recentlyRemoved` race handling, and the mirror-restoration behavior itself, including a test that it does *not* resurrect metadata after a layer is removed (the exact regression a sloppier version of this fix could introduce).

One thing worth a second look, not blocking: `syncLayers()` now attaches an observer to every layer's own node, not just `styleTargets` like the backdrop. Probably the right defensive call since any layer node could get its style clobbered the same way, but worth confirming that's intentional rather than broader than needed.

The bigger gap is that CI has never actually run on this commit — `total_count: 0` workflow runs via the Actions API. Looks like a maintainer needs to approve running workflows for this PR before there's any automated signal on lint/types/tests, separate from how it reads on manual review.

spokodev:
Intentional. `layer.node` is already a mirror target — `applyLayerStackMetadata(layer, index, layer.node)` writes `--layer-index`, `--nested-layer-count`, and `data-nested`/`data-has-nested` onto it on `main`, before the `styleTargets` loop. A framework re-render that replaces the node's inline `style` wipes those the same way it wipes the backdrop's, so the observer covers every element the metadata is written to. Restricting it to `styleTargets` would leave the layer node's own vars unprotected.

Issue #3149, opened 2026-05-27
Controlled `Drawer` flickers on swipe/backdrop close when `bind:open` setter has delay


adjsky:
### Description

When using controlled `Drawer` with `bind:open`, where setter is asynchronous (in my case it is history api, but for simplicity i use `setTimeout` in the example), backdrop and content start flickering when closing the drawer, but backdrop flickers only once, in the first try.

I was confused by this behavior because dialog in the same scenario does not flicker.

https://github.com/user-attachments/assets/0b010bf7-879e-4adf-b322-d08ef856d38f

### Link to Reproduction (or Detailed Explanation)

https://stackblitz.com/edit/47y7zurq-qdnq4knm?file=src%2FApp.svelte&showSidebar=0

### Steps to Reproduce

1. Click `Toggle Drawer`.
2. Click on backdrop.
3. See the flicker.
4. Click `Toggle Drawer` again.
5. Click on backdrop.
6. See no flicker.
7. Click `Toggle Drawer` again.
8. Try to swipe content.
9. See how content jumps.
10. Click `Toggle Dialog`.
11. Click on backdrop.
12. See no flicker.

### Ark UI Version

5.22.0

### Framework

- [ ] React
- [ ] Solid
- [ ] Vue
- [x] Svelte

segunadebayo:
I just pushed a fix for this. We'll release an update shortly in Ark UI

If the issue persists after upgrading, I'll re-open it.

adjsky:
The issue still seems to persist, backdrop still flickers when closing the drawer for the first time.

https://github.com/user-attachments/assets/e6f1e9f2-93fb-42cd-8688-46c45ffe3c0c

PR #3197, opened 2026-06-28
fix(drawer): prevent backdrop flickering when using an async open setter

adjsky:
Fixes the backdrop part in #3149.

Currently, `deferClearDragOffset` adds `animation: none !important` to the backdrop node, `clearSwipeOpenAnimation` is intended to clear this style, but due to a timing issue, it runs before the `CONTROLLED.CLOSE` event is received, and since the drawer is not yet in closing state, the opening animation runs.

## 🚀 New behavior

Now, if there is a defined `onOpenChange` prop, `clearSwipeOpenAnimation` only runs when the `CONTROLLED.CLOSE` event is received.

No.

### DIFF
--- a/.changeset/ninety-bushes-fold.md
+++ b/.changeset/ninety-bushes-fold.md
@@ -1,0 +1,5 @@
+---
+"@zag-js/drawer": patch
+---
+
+fix(drawer): prevent backdrop flickering when using an async open setter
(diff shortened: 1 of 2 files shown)


segunadebayo:
Thanks a lot for the fix and the clear reproduction, @adjsky! We addressed this on `main` in commit 6091040e0, taking a slightly different approach: instead of adding a new guard, we keep the swipe animation overrides in place until the drawer actually enters the `closing` state, so a controlled close driven by an async `open` setter no longer replays the enter animation while still open.

Closing since this is now resolved on `main`. Really appreciate the contribution!

PR #3202, opened 2026-07-05
feat(splitter): support data-collapsed attr

pedrodesu:
## 📝 Description

> Add `data-collapsed` data attribute to splitter's panels.

## 🚀 New behavior

> When a given splitter panel gets collapsed, it toggles the booleanish data attribute `data-collapsed`.

## 📝 Additional Information

This simple change can be very helpful to, for instance, control the styling of the panel with CSS only, based on whether it's collapsed or not. Overall it's behavior that should probably be exposed.

### DIFF
--- a/packages/docs/data/data-attr.json
+++ b/packages/docs/data/data-attr.json
@@ -1863,6 +1863,7 @@
       "data-part": "panel",
       "data-orientation": "The orientation of the panel",
       "data-dragging": "Present when in the dragging state",
+      "data-collapsed": "Present when collapsed",
       "data-id": "",
       "data-index": "The index of the item"
     },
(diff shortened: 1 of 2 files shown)


segunadebayo:
Thanks for this. Going to close it.

The CSS hook is useful, but `isPanelCollapsed(id)` already covers it. You can put that on the panel yourself:

```jsx
<div {...api.getPanelProps({ id })} data-collapsed={api.isPanelCollapsed(id) || undefined} />
```

We don't want to bake that into `getPanelProps`. Collapse isn't a machine state. It's a size check against `collapsedSize`, and that value can still be a guess before the root is measured. Putting it on the element makes a derived layout check look like first-class state, which it isn't.

Appreciate the thought.

Issue #3207, opened 2026-07-10
Stacked dialog exit animation hidden by layer z-index reset


Hwacc:
# 🐛 Bug report

When stacked dialogs are used, closing the top dialog can make its content exit animation visually disappear. The Presence state and CSS animation events still run correctly, but the closing dialog's `positioner` loses its layer CSS variables during `layerStack.remove()`, causing its z-index to collapse while the node is still mounted for the exit animation.

## 💥 Steps to reproduce

1. Using the official Zag.js nextjs examples -> Dialog examples -> Popover nested.
2. Ensure the Dialog component is wrapped with Presence so that the close animation can play.
3. Open D1 -> D2 -> D3
4. Observe that D3/D2 content exit animation is not visible, or appears hidden behind the lower dialog/backdrop.

## 🧐 Expected behavior

The closing top dialog should keep its resolved stacking order until its exit animation completes. Its content and backdrop should remain visible while `data-state="closed"` animations are running.

## 🧭 Possible Solution

Freeze the resolved z-index before clearing layer style metadata in `layerStack.remove()`.

```ts
layer.styleTargets?.forEach((getTarget) => {
  const target = getTarget()
  if (target) {
    const { zIndex } = getComputedStyle(target)
    if (zIndex && zIndex !== "auto") {
      target.style.setProperty("z-index", zIndex)
    }
    clearLayerStyleMirror(target)
  }
})
```

Then clear the frozen inline z-index when the layer is active again in `syncLayers()`:

```ts
target.style.removeProperty("z-index")
applyLayerStackMetadata(layer, index, target)
```

## 🌍 System information

| Software         | Version(s) |
| ---------------- | ---------- |
| Zag Version      | 1.42.0 |
| Browser          | Chrome |
| Operating System | Windows |

## 📝 Additional information

Please check the videos:

- Issue:
https://github.com/user-attachments/assets/bb04b80e-a823-4d41-9d32-c688a4172aa2

- solved:
https://github.com/user-attachments/assets/23431831-c761-4e5a-848c-6ee61badfaed

segunadebayo:
This issue has been automatically closed due to inactivity. If you believe this closure was in error, please reopen the issue or add a comment. Thanks for your understanding.

Note: This is an automated message.

PR #3208, opened 2026-07-10
fix(dismissable): Preserve target z-index when updating layer stack

Hwacc:
This PR is aimed to address #3207 

 💣 Is this a breaking change (Yes/No):  No


kalisaNkevin:
Traced through why this happens: Dialog registers its positioner as a style target, not the primary layer node (`layerStyleTargets: [getBackdropEl, getPositionerEl]` in `dialog.machine.ts`), and `remove()` only clears the style-mirror on style targets — the primary content node's own `--layer-index` is untouched. So the positioner's `z-index: var(--z-index)` rule goes dead the instant `--z-index` is stripped, collapsing to `auto` mid-exit-animation. That matches the reported symptom exactly.

Freezing the resolved z-index as a literal inline `z-index` (which wins on specificity over the now-broken `var()` rule) is the right fix, and the second commit's `removeProperty("z-index")` in `syncLayers` correctly handles the case where the same node re-enters the stack later — without it a stale frozen z-index would override the real one on reopen. Three commits, but they read as one coherent fix.

Two things still missing, same gap as #3177:

- No changeset. Based on how #3177 documented the same package, something like:
  ```
  ---
  "@zag-js/dismissable": patch
  ---

  Fix `z-index` collapsing on layer style targets (e.g. a dialog's `positioner`)
  when a layer is removed, which hid stacked dialogs' exit animations behind
  lower layers.
  ```
- No regression test. If #3177 lands first, there'll be a test file in this package to add one to; right now there isn't.

One sequencing note: this and #3177 both touch the same two `forEach` blocks in `remove()`/`syncLayers()`. No logical conflict — one freezes z-index, the other manages a `MutationObserver` — but whichever merges second will need a rebase.

CI-wise: same story as #3177. Checked the Actions API — there's exactly one workflow run (`Quality`), status `action_required`. Needs a maintainer to approve it before there's any real signal; the `unstable` mergeable state reflects that pending approval, not an actual conflict.

Hwacc:
I've updated the branch and filled in the missing test cases.  

As @kalisaNkevin said, this PR has a different starting point from #3177.  Here, the goal is to freeze the `z-index` during the remove phase, so that the stacking order remains correct while the exit animation is playing. In contrast, #3177 aims to keep metadata in sync using a `MutationObserver` at creation time.

What really bothers me, though, is this: as you can see in my earlier commit df48d395c1c74a37037d6c13019271ceb7a2c4fc, I originally planned to use a `WeakMap` to preserve and restore the `inline z-index` set by the user on each layer. But I later rolled back that change. I feel this approach runs counter to Zag's original design intent — leaving `--layer-index` to be handled by the consuming side's CSS.

MILLERMARRU (commented):
The three-layer test is what convinced me this is correctly scoped, it isolates the actual bug (removing the top of a stack re-syncs the lower layers via `syncLayers`, and without the freeze that re-sync was also touching the just-removed, still-animating node's stacking) from the simpler two-layer cases that wouldn't have caught it. Freezing the resolved z-index inline before clearing the layer vars is the right approach since it captures the value that was actually in effect at removal time, independent of the `calc(... var(--layer-index))` expression that produced it. The reactivation path correctly drops the frozen value via `removeProperty` before `applyLayerStackMetadata` re-applies live stacking, so a node getting reused for a new layer instance won't carry stale frozen state forward.

FrancoKaddour (commented):
This is a clean fix — freezing the resolved `z-index` before clearing the layer vars is exactly right for keeping the exiting layer visually on top through its close animation, and the three-layer test does a great job isolating the `syncLayers` re-sync bug from the trivial cases. Nice work.

One edge case I couldn't find covered, worth a sanity check: the frozen inline `z-index` is only dropped in `syncLayers` when the **same node is reactivated** (`removeProperty("z-index")` before re-applying metadata). What happens to a `styleTarget` that is *neither* reactivated *nor* unmounted after close? The common lifecycle is close → exit animation → unmount, so it's moot there. But if a target is a persistent/shared node (e.g. a backdrop or positioner reused outside a fresh `layerStack.add`), it'd keep the frozen inline `z-index` indefinitely and could shadow the stacking of whatever reuses that node later. Might be worth confirming all `styleTargets` are guaranteed to either unmount or go back through `add` — or, if not, clearing the freeze on a plain close too.

Minor, separate note: `getComputedStyle(target)` inside the `remove` loop forces a sync style flush per target. For a deep stack all closing in the same frame that's N reflows — probably negligible in practice, but if it ever shows up, caching the read or reading all targets before writing would avoid the layout thrash. Non-blocking.

Approach and test coverage both LGTM. 👍

PR #3209, opened 2026-07-10
fix(hover-card): keep open while pointer crosses gap between trigger and content

### LINKED ISSUE #3181: HoverCard closes when moving pointer across gutter gap between trigger and content
Hi, thanks for the great work on Ark UI/Zag.

I'm using `@ark-ui/react@5.37.2`, which depends on `@zag-js/hover-card@1.41.2`.

I noticed an interaction issue with `HoverCard` when using `positioning.gutter`. The gutter visually looks good because it creates spacing between the trigger and the floating content, but when the user moves the pointer from the trigger to the content through that gap, the hover card closes because the pointer briefly leaves both the trigger and content.

## Reproduction

```tsx
import { HoverCard } from "@ark-ui/react"

export function Example() {
  return (
    <HoverCard.Root
      openDelay={10}
      closeDelay={10}
      positioning={{ placement: "bottom-start", gutter: 8 }}
    >
      <HoverCard.Trigger>
        User
      </HoverCard.Trigger>

      <HoverCard.Positioner>
        <HoverCard.Content>
          <a href="/profile">Profile</a>
          <a href="/settings">Settings</a>
        </HoverCard.Content>
      </HoverCard.Positioner>
    </HoverCard.Root>
  )
}
Current behavior
When moving the mouse from the trigger to the content, the pointer crosses the 8px gutter. During that moment, the pointer is not over the trigger or the content, so the hover card closes.
This makes gutter difficult to use for interactive hover cards unless closeDelay is increased.
Expected behavior
There should be a way to preserve the visual gutter while allowing the user to safely move the pointer from the trigger to the content.
Possible solutions could be:
- Add support for a "safe polygon" style interaction similar to Floating UI's safePolygon.
- Add an option to treat the gutter area as an interactive hover bridge.
- Expose a prop or recommended pattern for creating a hover-safe gap without relying on closeDelay.
Why increasing closeDelay is not ideal
Increasing closeDelay helps, but it changes the close behavior globally. It makes the hover card feel less responsive when the user intentionally moves away. In this case the desired behavior is only to tolerate the physical gap between trigger and content, not delay all closes.
Environment
- @ark-ui/react: 5.37.2
- @zag-js/hover-card: 1.41.2
- React: 19.2.4
- Next.js: 16.2.7
Thanks!

Arman-Luthra:
Closes #3181

## 📝 Description

When `positioning.gutter` (or an `offset`) creates a visual gap between the trigger and the content, moving the pointer across that gap can close the hover card, because the pointer briefly hovers neither element. This PR adds a grace area (safe polygon) between the trigger and the content, similar to Floating UI's `safePolygon` and Radix's hover card.

- On `pointerleave`, the machine enters `closing` and closes after `closeDelay`, regardless of where the pointer is heading.
- Crossing the gutter gap works only if the crossing takes less than `closeDelay`, so users are forced to increase `closeDelay` globally to make interactive hover cards usable with a gutter.

## 🚀 New behavior

- When the pointer leaves the trigger towards the content (or leaves the content towards the trigger), a grace area polygon is computed from the exit point and the target element's rect (reusing `getElementPolygon`/`isPointInPolygon` from `@zag-js/rect-utils`, the same primitives the menu machine uses for its submenu intent polygon).
- While the pointer stays inside the grace area, the `CLOSE_DELAY` expiry is ignored, so the card stays open during the traversal no matter how small `closeDelay` is.
- As soon as the pointer exits the grace area (or leaves the document), the `closing` state re-enters and the regular `closeDelay` countdown restarts — so moving away still closes the card with the usual delay.
- The grace area is only created when the pointer actually exits on the side facing the target element. Leaving in any other direction (e.g. jumping to an unrelated part of the page) keeps the previous close behavior.

Interaction with `closeDelay`: the delay is effectively paused while the pointer is in transit between the two elements, and restarts once the pointer leaves the corridor. Entering the content/trigger cancels the close as before.

No. Public API is unchanged; the grace area is internal (`refs`). The only new dependency is the existing workspace package `@zag-js/rect-utils`.

## 📝 Additional Information

- Added three e2e tests: slow traversal trigger → content and content → trigger through the gutter (both fail on `main`, pass with this fix), plus a test asserting the card still closes when the pointer enters the gap and then veers away.
- Limitation (matches Radix behavior): if the pointer stops inside the gap and never moves again, the card stays open until the pointer moves out of the grace area. Leaving the document while in the gap is treated as exiting the grace area.
- The grace area is intentionally not created for diagonal exits through a non-facing side (e.g. leaving the trigger's right edge when the content sits below); those fall back to the current `closeDelay` behavior.

### DIFF
--- a/.changeset/short-lamps-brake.md
+++ b/.changeset/short-lamps-brake.md
@@ -1,0 +1,7 @@
+---
+"@zag-js/hover-card": patch
+---
+
+Fix issue where hover card closes when the pointer crosses the gap between the trigger and the content (e.g. when using
+`positioning.gutter`). The close delay is now paused while the pointer travels across the grace area between both
+elements.
--- a/e2e/hover-card.e2e.ts
+++ b/e2e/hover-card.e2e.ts
@@ -97,3 +97,76 @@ test("should remain open after moving from content back to trigger", async ({ pa
   await page.hover(trigger)
   await expect(page.locator(content)).toBeVisible()
 })
+
+test("should remain open when pointer crosses the gutter gap slowly (trigger to content)", async ({ page }) => {
+  await page.hover(trigger)
+  await page.waitForSelector(content)
+  await expect(page.locator(content)).toBeVisible()
+
+  const triggerBox = await page.locator(trigger).boundingBox()
+  const contentBox = await page.locator(content).boundingBox()
+  if (!triggerBox || !contentBox) throw new Error("missing bounding boxes")
+
+  // walk straight down from the trigger into the gutter gap, pausing between
+  // steps so the traversal takes longer than the default `closeDelay` (300ms)
+  const x = triggerBox.x + triggerBox.width / 2
+  const startY = triggerBox.y + triggerBox.height - 1
+  const endY = contentBox.y + 1
+
+  await page.mouse.move(x, startY)
+  const steps = Math.max(6, Math.ceil(Math.abs(endY - startY) / 2))
+  for (let i = 1; i <= steps; i++) {
+    await page.mouse.move(x, startY + ((endY - startY) * i) / steps)
+    await page.waitForTimeout(100)
+    expect(await page.locator(content).isVisible()).toBe(true)
+  }
+})
+
+test("should close when pointer leaves the grace area between trigger and content", async ({ page }) => {
+  await page.hover(trigger)
+  await page.waitForSelector(content)
+  await expect(page.locator(content)).toBeVisible()
+
+  const triggerBox = await page.locator(trigger).boundingBox()
+  const contentBox = await page.locator(content).boundingBox()
+  if (!triggerBox || !contentBox) throw new Error("missing bounding boxes")
+
+  // move into the gutter gap and linger past the default `closeDelay` (300ms)
+  const x = triggerBox.x + triggerBox.width / 2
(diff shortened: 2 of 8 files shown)


kalisaNkevin:
Took a closer look at this one. The fix and tests hold up: the default gutter (8px, from `@zag-js/popper`'s `get-placement.ts:16`) is enough for the original bug to reproduce, and the new e2e tests wait longer than the default 300ms `closeDelay` (≥600ms of traversal), so they genuinely fail on `main` and pass here — not just timing luck. The grace-area approach also isn't a new pattern here; it reuses `getElementPolygon`/`isPointInPolygon` from `@zag-js/rect-utils`, the same primitives `menu.utils.ts` already uses for submenu intent detection.

Two small nits, nothing blocking:

- The JSDoc on `createGraceArea` (`hover-card.utils.ts:15-24`) has four `@param` lines that just restate the parameter names — worth trimming, since nothing else nearby (`getElementPolygon`, `menu.utils.ts`'s `isWithinPolygon`) carries JSDoc like this.
- The comment in `onPointerLeave` ("pointer left the document...") repeats almost the same wording as the comment on `GRACE_AREA.EXIT` a few lines away — could just say it once.

Looks good to merge otherwise.

segunadebayo:
This is resolved in Zag v2. We'll release it shortly

PR #3210, opened 2026-07-13
fix(date-input): commit deferred entry on segment leave

midub:
## Summary

Editing over a committed date showed the stale value until the whole control lost focus.

**Repro:** commit `6/2/2000`, refocus, type `6`, `3`, Tab → segments kept rendering `6/2/2000` instead of `6/3/2000` until clicking outside. The day segment DOM even disagreed with itself: `data-value="3"` / `aria-valuenow="3"` but rendered text `2`.

**Root cause:** a partial segment entry (day `3` that could still become `30`) keeps `enteredKeys` pending, which defers the commit in `setSegmentValue`. The `segments` computed then prefers the committed value over the in-progress display value (`isFullyCommitted` branch), so the stale date stays rendered. Only control blur (`confirmPlaceholder`) committed the deferred entry.

**Fix:** new `commitDeferredSegmentValue` action commits the pending entry when its segment is left via click, Tab, or arrow navigation (`SEGMENT.FOCUS`, `SEGMENT.ARROW_LEFT`, `SEGMENT.ARROW_RIGHT`) — the same semantics blur already applied, just at the moment the entry becomes final.

Also included: `fix(shared)` pins the examples nav sort to `en-US` collation. The sort ran with the OS locale on the server and the browser locale on the client; where collations disagree (e.g. Czech orders "Ch" after "H", flipping Checkbox/Clipboard) the nav failed hydration and the Next dev overlay blocked pointer events during local e2e runs on non-English machines.

## Test plan

- New e2e regression test `[editingValue] re-typing over a committed date shows the new digits before blur` — fails without the machine fix (`Expected "3", Received "2"`), passes with it.
- Full `date-input` e2e suite: 83/86 pass; the 3 failures are pre-existing local flakes (fail identically without this change; timing race between `type()` and rAF segment focus, green in CI).
- Changeset included (`@zag-js/date-input` patch).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.changeset/stale-segment-display-fix.md
+++ b/.changeset/stale-segment-display-fix.md
@@ -1,0 +1,7 @@
+---
+"@zag-js/date-input": patch
+---
+
+Fix stale value shown when editing over a committed date. A partial segment entry (e.g. day `3` that could still
+become `30`) kept the previous committed value rendered after moving to another segment until the whole control lost
+focus. The pending entry is now committed when its segment is left via click, Tab, or arrow navigation.
--- a/e2e/date-input.e2e.ts
+++ b/e2e/date-input.e2e.ts
@@ -464,6 +464,39 @@ test.describe("date-input [single]", () => {
     await I.seeSegmentIsNotPlaceholder("day")
   })
 
+  test("[editingValue] re-typing over a committed date shows the new digits before blur", async () => {
+    // Commit 6/2/2000: "6" completes the month (auto-advance), "2" is a partial day entry,
+    // Tab moves to year, "2000" completes the date.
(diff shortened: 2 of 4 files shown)


segunadebayo:
Thanks for the investigation and PR — really appreciate the time you put into this.

We've fixed the root cause on main by formatting segment text from the in-progress `displayValues` instead of committing deferred entries on segment leave: https://github.com/chakra-ui/zag/commit/2bf453c41

Closing in favor of that approach.

midub:
Your solution is better, thank you.

PR #3227, opened 2026-07-25
fix(hover-card): keep open while trigger remains focused

### LINKED ISSUE #3224: HoverCard closes on pointer leave while the trigger remains focused
# 🐛 Bug report

A `HoverCard` closes after the pointer leaves its trigger, even though the trigger remains focused.

## 💥 Steps to reproduce

1. Click “Start here”.
2. Press `Tab` to focus “Profile” and open the card.
3. Move the pointer over “Profile”, then move it away from both “Profile” and the card.
4. Wait 300 ms.
5. `Profile focus` still reads `focused`, while `HoverCard` changes to `closed`.

## 💻 Link to reproduction

StackBlitz reproduction: https://stackblitz.com/edit/zag-hover-card-pointer-leave-repro?file=src%2Fmain.jsx

The same issue reproduces in the [Ark UI Hover Card example](https://ark-ui.com/docs/components/hover-card).

## 🧐 Expected behavior

A `HoverCard` opened by focus should remain open while its trigger remains focused, even after the pointer enters and leaves the trigger.

## 🧭 Possible Solution

`POINTER_LEAVE` currently enters the closing state without considering focus. Closing may need to wait until neither pointer hover nor trigger focus remains.

## 🌍 System information

| Software         | Version(s)                  |
| ---------------- | --------------------------- |
| Zag Version      | `@zag-js/hover-card` 1.42.0 |
| Browser          | Chromium 150                |
| Operating System | macOS 15.7.3                |

## 📝 Additional information

The relevant transition is in [hover-card.machine.ts](https://github.com/chakra-ui/zag/blob/2b199781b1a8d9b8a57b03dff443cf5dbf55c912/packages/machines/hover-card/src/hover-card.machine.ts#L154-L166).

Focus behavior appears intentional. The E2E tests cover opening on focus and closing on blur:

https://github.com/chakra-ui/zag/blob/2b199781b1a8d9b8a57b03dff443cf5dbf55c912/e2e/hover-card.e2e.ts#L34-L78

PR #316 fixed the inverse interaction: a card opened by focus should remain open after blur while the pointer is still over it:

https://github.com/chakra-ui/zag/pull/316

The docs currently say “Opens on hover only”, which appears inconsistent with the implementation and tests:

https://github.com/chakra-ui/zag/blob/2b199781b1a8d9b8a57b03dff443cf5dbf55c912/website/data/components/hover-card.mdx#L13-L20

This is separate from #3181 and PR #3209, which cover pointer movement across the gap between the trigger and content.

aose-yuu:
Closes #3224

## 📝 Description

Keep a hover card open when the pointer leaves a trigger that still has keyboard focus.

Moving the pointer away starts closing the hover card even when its trigger remains focused.

## 🚀 New behavior

Pointer leave no longer closes the hover card while the trigger is focused. Pointer state is cleared so the card still closes when focus later leaves the trigger, and an end-to-end test covers both parts of the interaction.

No.

## 📝 Additional Information

#3209 changes the same pointer-leave transition, so this branch may need a small rebase if that PR lands first.

### DIFF
--- a/.changeset/hover-card-focus-pointer-leave.md
+++ b/.changeset/hover-card-focus-pointer-leave.md
@@ -1,0 +1,5 @@
+---
+"@zag-js/hover-card": patch
+---
+
+Keep the hover card open when the pointer leaves a trigger that remains focused.
--- a/e2e/hover-card-multiple-trigger.e2e.ts
+++ b/e2e/hover-card-multiple-trigger.e2e.ts
@@ -61,6 +61,24 @@ test.describe("hover-card / multiple triggers", () => {
     await expect(page.locator(content)).toContainText("Alice Johnson")
   })
 
+  test("should remain open after pointer leaves the focused active trigger", async ({ page }) => {
+    await page.waitForLoadState("networkidle")
+    await page.locator(trigger(2)).focus()
+    await expect(page.locator(trigger(2))).toBeFocused()
+    await expect(page.locator(content)).toBeVisible()
+    await expect(page.locator(content)).toContainText("Bob Smith")
+
+    await page.hover(trigger(2))
+    await page.mouse.move(0, 0)
+    await page.waitForTimeout(500)
+    await expect(page.locator(trigger(2))).toBeFocused()
+    await expect(page.locator(content)).toBeVisible()
+    await expect(page.locator(content)).toContainText("Bob Smith")
+
+    await page.locator(trigger(2)).evaluate((element) => element.blur())
+    await expect(page.locator(content)).toBeHidden()
+  })
+
   test("should close hover card on escape", async ({ page }) => {
     await page.hover(trigger(1))
     await expect(page.locator(content)).toBeVisible()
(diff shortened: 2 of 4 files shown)


kalisaNkevin:
Nice fix, and pretty minimal. The clever bit is it doesn't need a new closing path — clearing `isPointer` on the focused-trigger `POINTER_LEAVE` is enough, because the `open` state's existing `TRIGGER_BLUR` handler already checks `not(isPointer)` to decide whether to close. So once focus actually leaves the trigger later, that existing logic just takes over and closes it normally. No new states, no extra bookkeeping.

I also checked how this holds up for the multi-trigger setup, since the guard reads `context.get("triggerValue")` to resolve the *active* trigger rather than just checking "is anything focused" — so it should only suppress closing for the specific trigger currently driving the open card, not any trigger on the page. That looks right by inspection, though there's no test covering the multi-trigger + focus combo specifically (`hover-card-multiple-trigger.e2e.ts` wasn't touched), so worth a manual sanity check there.

Changeset and e2e test are both present and read fine. Same CI situation as the last three though — one workflow run, `action_required`, so nothing's actually been verified yet until a maintainer approves it.

aose-yuu:
I added coverage for this in [18af8fb](https://github.com/aose-yuu/zag/commit/18af8fbab53cde189229a8b65dc6b0dd3152cd03).

The new test focuses the second trigger, moves the pointer away, and checks that the card stays open.
It also checks that the card closes once the trigger loses focus.

PR #3233, opened 2026-07-27
fix(marquee): use content size for animation duration so speed is viewport-independent

### LINKED ISSUE #3231: @zag-js/marquee: Actual speed depends on content width when content is smaller than viewport
# 🐛 Bug report

The marquee's actual scrolling speed depends on the content width whenever the content is smaller than the viewport.

The machine calculates the duration using `rootSize`. However, every content element is translated by only `100%` of its own size, so the actual animation distance is `contentSize`.

As a result, the configured `speed` only represents the actual pixel speed when the content is at least as large as the viewport.

## 💥 Steps to reproduce

1. Create a marquee with `speed: 100`.
2. Render the content using `api.contentCount` and `api.getContentProps()`.
3. Use a `1000px`-wide viewport and content that is `100px` wide.
4. Observe that `--marquee-duration` is `10s`, while the content moves only `100px`. The actual speed is therefore `10px/s` instead of `100px/s`.
5. Replace the content with an element at least as wide as the viewport and observe that it moves at approximately `100px/s`.

For example, with the following measurements:

```
rootSize: 1000px
contentSize: 100px
speed: 100px/s
duration: 10s
```

The element moves only `100px` over `10s`, resulting in an actual speed of `10px/s` instead of `100px/s`.

## 💻 Link to reproduction

CodeSandbox reproduction: https://codesandbox.io/p/devbox/7p3922

## 🧐 Expected behavior

The `speed` property should represent a consistent number of pixels per second regardless of:

- Content width
- Viewport width

With `speed: 100`, every content element should move at approximately `100px/s`.

## 🧭 Possible Solution

The duration should be based on the actual animation distance.

Currently, every content element is translated by `100%` of its own size, so the animation distance is always `contentSize`. However, `calculateDuration` uses `rootSize` when the content is smaller than the viewport.

This causes the actual pixel speed to depend on the content-to-viewport ratio.

The duration calculation could therefore be simplified to:

```ts
function calculateDuration(options: { contentSize: number; speed: number }): number {
  return options.contentSize / options.speed
}
```

Alternatively, the animation distance would need to be changed to match the existing duration calculation.

## 🌍 System information

| Software         | Version(s)                         |
| ---------------- | ---------------------------------- |
| Zag Version      | @zag-js/marquee@^1.42.0             |
| Browser          | Microsoft Edge WebView2 / Chromium |
| Operating System | Windows                            |

## 🧢 Your Company/Team

N/A

## 📝 Additional information

Relevant source locations:

- [`marquee.machine.ts`](https://github.com/chakra-ui/zag/blob/main/packages/machines/marquee/src/marquee.machine.ts)
- [`marquee.connect.ts`](https://github.com/chakra-ui/zag/blob/main/packages/machines/marquee/src/marquee.connect.ts)
- [`marquee.utils.ts`](https://github.com/chakra-ui/zag/blob/main/packages/machines/marquee/src/marquee.utils.ts)

The issue is also present in the current `main` branch.

FrancoKaddour:
Closes #3231

The marquee computes `--marquee-duration` from `rootSize`, but each content element is translated by `100%` of its own size (`translateX(100%)` in the `marqueeX`/`marqueeY` keyframes), so the actual animation distance is always `contentSize`. When the content is smaller than the viewport, the configured `speed` no longer matches the real pixel speed — it scales by `contentSize / rootSize`. With the reporter's example (1000px viewport, 100px content, `speed: 100`), the duration was `10s` for a 100px move → 10px/s instead of 100px/s.

**Fix:** in the non-`autoFill` branch, base the duration on `contentSize` (the real translation distance) rather than `rootSize`. `calculateDuration` is moved to `marquee.utils.ts` and unit-tested (matching the utils-test convention used across the machines).

Verified: the new `marquee.utils.test.ts` covers the reported case and asserts `contentSize / duration === speed` across viewport/content ratios; the bug cases fail against the previous `rootSize`-based logic.

### DIFF
--- a/.changeset/marquee-duration-content-size.md
+++ b/.changeset/marquee-duration-content-size.md
@@ -1,0 +1,7 @@
+---
+"@zag-js/marquee": patch
+---
+
+Fix the marquee scrolling speed depending on the content width. The duration is now derived from the
+content size (the actual translation distance) instead of the root size, so the configured 
+matches the real pixel speed even when the content is smaller than the viewport.
--- a/packages/machines/marquee/src/marquee.machine.ts
+++ b/packages/machines/marquee/src/marquee.machine.ts
@@ -1,5 +1,6 @@
 import { createMachine } from "@zag-js/core"
 import { dom } from "./marquee.dom"
+import { calculateDuration } from "./marquee.utils"
 import type { MarqueeSchema } from "./marquee.types"
 
 export const machine = createMachine<MarqueeSchema>({
@@ -208,19 +209,3 @@ export const machine = createMachine<MarqueeSchema>({
     },
   },
 })
-
-function calculateDuration(options: {
-  rootSize: number
-  contentSize: number
-  speed: number
-  multiplier: number
-  autoFill: boolean
-}): number {
-  const { rootSize, contentSize, speed, multiplier, autoFill } = options
-
-  if (autoFill) {
-    return (contentSize * multiplier) / speed
-  }
-
-  return contentSize < rootSize ? rootSize / speed : contentSize / speed
-}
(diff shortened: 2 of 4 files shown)


FrancoKaddour:
Thanks for the quick merge, @segunadebayo! 🙌

PR #3240, opened 2026-07-29
fix(image-cropper): account for flip when mapping crop box to source pixels

### LINKED ISSUE #3238: [image-cropper] getCroppedImage() returns the wrong region when the image is flipped or rotated
# 🐛 Bug report

`getCroppedImage()` (and `getCropData()`) return the wrong region of the source image whenever the image is flipped or rotated. Zoom and pan alone behave correctly.

The crop-box → source-pixels mapping in `getCropSourceRect` (image-cropper.utils.ts) inverts only the pan offset and zoom — flip and rotation are not parameters of the function. The preview image, however, is CSS-transformed with `translate(offset) rotate(rotation) scale(zoom × flipSign)` (see `getImageProps`), so whenever flip or rotation is active, the sampled region no longer corresponds to what's inside the crop box on screen. `drawCroppedImageToCanvas` then applies `ctx.rotate(...)` / `ctx.scale(±1, ±1)` to the already-wrong tile, producing e.g. a mirrored copy of the mirror-opposite region on horizontal flip.

## 💥 Steps to reproduce

The reproduction uses a 1200×800 test image with four solid quadrants (TL red, TR green, BL blue, BR yellow) in a 600×400 viewport, with the crop box over the top-right area of the screen.

1. Open the sandbox below
2. Click **Flip horizontally** — the crop box now visibly contains **red** content (the mirrored top-left quadrant)
3. Click **getCroppedImage()** — the rendered result is **green** (the top-right quadrant, mirrored about the image centre), not the red content shown inside the crop box
4. Click **Reset**, then **Rotate 90°**, then **getCroppedImage()** — also returns content that doesn't match the crop box

## 💻 Link to reproduction

CodeSandbox reproduction: https://codesandbox.io/s/pvvc92

## 🧐 Expected behavior

`getCroppedImage()` returns exactly the pixels visible inside the crop box, for any combination of pan, zoom, rotation, and flip.

## 🧭 Possible Solution

For **flip** the inverse mapping is an exact sign fix inside `getCropSourceRect`:

```
sourceCenterX = imageCenterX + ((cropCenterX - viewportCenterX - offset.x) / zoom) * flipSignX * scaleX
sourceCenterY = imageCenterY + ((cropCenterY - viewportCenterY - offset.y) / zoom) * flipSignY * scaleY
```

**Rotation** can't be expressed as an axis-aligned source rect. A robust approach that handles all four transforms at once is to stop computing a source rect and instead draw the full image through the same affine transform the CSS applies, letting canvas clipping produce the crop. Verified pixel-exact in-browser for flip, rotation (15°/45°/90°/180°), zoom, pan, and combinations:

```js
// output at source resolution rather than viewport CSS pixels
const bx = viewportRect.width / naturalSize.width
const by = viewportRect.height / naturalSize.height
const kx = 1 / (zoom * bx)
const ky = 1 / (zoom * by)

canvas.width = Math.round(crop.width * kx)
canvas.height = Math.round(crop.height * ky)

ctx.scale(kx, ky)
ctx.translate(-crop.x, -crop.y)
ctx.translate(viewportRect.width / 2 + offset.x, viewportRect.height / 2 + offset.y)
ctx.rotate((rotation * Math.PI) / 180)
ctx.scale(zoom * (flip.horizontal ? -1 : 1), zoom * (flip.vertical ? -1 : 1))
ctx.scale(bx, by)
ctx.translate(-naturalSize.width / 2, -naturalSize.height / 2)
ctx.drawImage(imageEl, 0, 0)
```

This would also fix two side effects of the current implementation:

- the output canvas is sized in viewport CSS pixels (`canvas.width = crop.width`), so crops are downsampled to screen resolution instead of keeping the source image's resolution;
- under rotation, rotating the axis-aligned tile about the canvas centre leaves transparent wedges in the output corners even when the on-screen crop box is fully covered by the image.

## 🌍 System information

| Zag Version      | 1.41.2 and 1.42.0 (crop source identical); also present on `main` and `v2` HEAD, and in 2.0.0-next.0 |
| Browser          | Chrome (not browser-specific — the mapping maths is off) |
| Operating System | macOS |

## 📝 Additional information

Round-trip check: take a source pixel, compute where the CSS transform in `getImageProps` puts it on screen, centre a crop box there, and ask `getCropSourceRect` which source pixel is at the box centre. Source pixel (300, 200) in a 1200×800 image, 600×400 viewport:

| State | Maps back to | |
| --- | --- | --- |
| plain / zoomed 1.5× / panned / zoomed+panned | (300, 200) | ✅ correct |
| flipped horizontally | (900, 200) | ❌ mirrored in x |
| flipped vertically | (300, 600) | ❌ mirrored in y |
| rotated 15° | (362, 129) | ❌ |
| rotated 90° | (800, 100) | ❌ |
| rotated 180° | (900, 600) | ❌ |
| flipped h + zoomed 1.5× + panned | (900, 200) | ❌ mirrored in x |

Context: `getCropSourceRect` was introduced in 5e602cb9cb (the fix for #3150, shipped in 1.41.2). That commit fixed the viewport→natural scale conversion, but the new mapping never received flip/rotation. `getCropData()` uses the same mapping, so its rect is inconsistent with the on-screen selection under rotation as well.

FrancoKaddour:
Closes #3238

## 📝 Description

`getCroppedImage()` and `getCropData()` return the wrong region of the source image when a horizontal or vertical flip is active. Zoom and pan alone are fine.

## ⛳️ Current behavior

The preview is transformed with `translate(offset) rotate(rotation) scale(zoom × flipSign)` (see `getImageProps`), but `getCropSourceRect` — which maps the crop box back to source pixels — only inverts pan and zoom. Flip is not a parameter, so the sampled center is never mirrored: a horizontal flip of a crop over the top-right of the viewport samples the top-*right* quadrant, and the canvas's `ctx.scale(±1)` then mirrors that already-wrong tile — the top-*left* quadrant the user actually sees is never sampled.

## 🚀 New behavior

`getCropSourceRect` now takes the `flip` state and mirrors the sampled region's center across the image center per axis, leaving the sampled size and the viewport-space pan offset untouched. Both consumers (`getCropData` and the canvas draw) pass it through; the canvas's existing `ctx.scale(flip)` still supplies the visual orientation.

Added unit tests for horizontal / vertical / combined flips, the centered-crop invariant, the pan-offset interaction, and a regression matching the reported reproduction. Reverting the fix fails the mirroring assertions while the existing zoom/pan/scale tests stay green.

**Note on rotation:** the title also mentions rotation. Flip is a clean, exact fix (an axis-aligned source rect stays axis-aligned under a mirror). Arbitrary rotation (`rotation` is clamped 0–360, not quantized to 90°) is deeper: an axis-aligned source rect can't represent a rotated crop, so a fully-correct `getCroppedImage()` under rotation needs `drawCroppedImageToCanvas` to replicate the full CSS transform via `ctx.setTransform(...)`, and `getCropData()`'s axis-aligned rect is ambiguous under rotation. I scoped this PR to the flip fix that resolves the reported reproduction, and I'm happy to follow up on rotation in a separate PR — I have the `setTransform` approach worked out.

No. `flip` is optional; existing behavior (no flip) is unchanged.

### DIFF
--- a/.changeset/image-cropper-flip-source-rect.md
+++ b/.changeset/image-cropper-flip-source-rect.md
@@ -1,0 +1,6 @@
+---
+"@zag-js/image-cropper": patch
+---
+
+Fix `getCroppedImage()` and `getCropData()` sampling the mirror-opposite region of the source image when a horizontal or
+vertical flip is active
--- a/packages/machines/image-cropper/src/image-cropper.connect.ts
+++ b/packages/machines/image-cropper/src/image-cropper.connect.ts
@@ -128,6 +128,7 @@ export function connect<T extends PropTypes>(
         offset,
         viewportSize: viewportRect,
         naturalSize,
+        flip,
       })
 
       return {
--- a/packages/machines/image-cropper/src/image-cropper.dom.ts
+++ b/packages/machines/image-cropper/src/image-cropper.dom.ts
@@ -53,6 +53,7 @@ export function drawCroppedImageToCanvas(params: ImageCropperService): HTMLCanva
     offset,
     viewportSize: viewportRect,
     naturalSize,
+    flip,
   })
 
   ctx.drawImage(
--- a/packages/machines/image-cropper/src/image-cropper.utils.ts
+++ b/packages/machines/image-cropper/src/image-cropper.utils.ts
@@ -491,19 +491,28 @@ interface SourceRectParams {
   offset: Point
   viewportSize: Size
   naturalSize: Size
+  flip?: FlipState | undefined
 }
 
 /**
  * Maps the crop rect (viewport coordinates) to the source rect in natural
  * image pixels. The viewport -> natural scale is `naturalSize / viewportSize`.
+ *
+ * The preview image is transformed with `translate() scale(zoom * flipSign)`,
+ * so a horizontal/vertical flip mirrors the sampled region across the image
+ * center. The offset (pan) is applied in viewport space and is unaffected by
+ * the flip.
  */
 export function getCropSourceRect(params: SourceRectParams): Rect {
-  const { crop, zoom, offset, viewportSize, naturalSize } = params
+  const { crop, zoom, offset, viewportSize, naturalSize, flip } = params
 
   const scaleX = viewportSize.width > 0 ? naturalSize.width / viewportSize.width : 1
   const scaleY = viewportSize.height > 0 ? naturalSize.height / viewportSize.height : 1
 
   const safeZoom = zoom > 0 ? zoom : 1
+
+  const flipX = flip?.horizontal ? -1 : 1
+  const flipY = flip?.vertical ? -1 : 1
 
   const viewportCenterX = viewportSize.width / 2
   const viewportCenterY = viewportSize.height / 2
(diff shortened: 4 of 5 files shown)


jamesofficer:
I think @segunadebayo may have already fixed this in commit: https://github.com/chakra-ui/zag/commit/3236a21325663ece63b2990a83c41bb920bcf356 - but correct me if I'm wrong!

Appreciate this PR either way!

FrancoKaddour:
Thanks @jamesofficer — you're right. @segunadebayo's `3236a21` (already on `main`) resolves this comprehensively: it reworks the crop/transform pipeline so `getCropData()` returns exact natural-image `corners` + `outputSize` and `getCroppedImage()` exports at natural resolution (with a new `maxSize` option), covering **both** flip and rotation — including the rotation case I'd flagged here as follow-up — plus an e2e regression with the quadrant image. That fully supersedes this PR, so I'll close it. Appreciate the heads-up, and nice work 🙌

PR #3246, opened 2026-08-03
fix(floating-panel): keep panel stack in sync on close and apply stack order to the positioner

### LINKED ISSUE #3242: Focusing a Floating Panel does not bring it to the front
# 🐛 Bug report

When two Floating Panels overlap, focusing the panel underneath updates its stack state, but the panel stays behind the one rendered later.

## 💥 Steps to reproduce

1. Open the StackBlitz reproduction.
2. Click **Focus A**.
3. Confirm that Panel A shows `data-topmost` and Panel B shows `data-behind`.
4. Panel B still covers Panel A in the overlapping area.

## 💻 Link to reproduction

StackBlitz reproduction: https://stackblitz.com/edit/zag-floating-panel-stacking-repro?file=src%2FApp.jsx

## 🧐 Expected behavior

The focused panel should be displayed above the other open panels.

## 🧭 Possible Solution

`trackPanelStack` sets `--z-index` on the Content element. Since each Positioner is a sibling stacking context, the Content z-index cannot change the order between panels.

The stack index may also need to be exposed on the Positioner.

## 🌍 System information

| Software         | Version(s)                      |
| ---------------- | ------------------------------- |
| Zag Version      | `@zag-js/floating-panel` 1.43.0 |
| Browser          | Chromium                        |
| Operating System | macOS                           |

## 📝 Additional information

After clicking **Focus A**:

- Panel A Content has `data-topmost` and `--z-index: 2`.
- Panel B Content has `data-behind` and `--z-index: 1`.
- Both Positioners remain at `z-index: 100`.

Panel B therefore stays above Panel A because it appears later in the DOM.

The stack index is currently written to Content in [`trackPanelStack`](https://github.com/chakra-ui/zag/blob/2fa60075845ea2e31e8b4da9c7ea797ff193fa0a/packages/machines/floating-panel/src/floating-panel.machine.ts#L314-L324). The [Positioner props](https://github.com/chakra-ui/zag/blob/2fa60075845ea2e31e8b4da9c7ea797ff193fa0a/packages/machines/floating-panel/src/floating-panel.connect.ts#L80-L94) do not expose that value.

### LINKED ISSUE #3243: Closing a Floating Panel does not update the topmost state
# 🐛 Bug report

Closing the topmost Floating Panel leaves its stack state unchanged. The closed panel keeps `data-topmost`, while the next open panel keeps `data-behind`.

## 💥 Steps to reproduce

1. Open the StackBlitz reproduction.
2. Click **Open A**.
3. Click **Open B**.
4. Close Panel B.
5. Panel A is still marked as `data-behind`.

## 💻 Link to reproduction

StackBlitz reproduction: https://stackblitz.com/edit/zag-floating-panel-stale-topmost-repro?file=src%2FApp.jsx

## 🧐 Expected behavior

Panel A should receive `data-topmost` when Panel B closes.

## 🧭 Possible Solution

`trackPanelStack` removes a panel from `panelStack` only when the machine is destroyed. It does not remove the panel when it closes.

Removing the panel on close would update `isTopmost` for the remaining panels. Opening it again would add it back through the existing open-state action.

## 🌍 System information

| Software         | Version(s)                      |
| ---------------- | ------------------------------- |
| Zag Version      | `@zag-js/floating-panel` 1.43.0 |
| Browser          | Chromium                        |
| Operating System | macOS                           |

## 📝 Additional information

After Panel B closes:

- Panel A is visible with `data-behind`.
- Panel B is hidden with `data-topmost`.
- Focusing Panel A updates the state and makes it topmost.

The panel is added to the stack when it opens, but removed only when `trackPanelStack` is cleaned up:

- [Open state](https://github.com/chakra-ui/zag/blob/2fa60075845ea2e31e8b4da9c7ea797ff193fa0a/packages/machines/floating-panel/src/floating-panel.machine.ts#L159-L178)
- [`trackPanelStack`](https://github.com/chakra-ui/zag/blob/2fa60075845ea2e31e8b4da9c7ea797ff193fa0a/packages/machines/floating-panel/src/floating-panel.machine.ts#L314-L329)

#3242 covers a different part of panel stacking. In that issue, the stack index is correct but does not affect the Positioner. Here, the stack state itself is stale after a panel closes.

FrancoKaddour:
Closes #3243
Closes #3242

## Problem

Two related panel-stacking bugs when multiple floating panels are open:

1. **Stale stack on close (#3243):** panels are added to the shared `panelStack` when they open, but only removed when the machine is *destroyed*. Closing the topmost panel left it in the stack, so it kept `data-topmost` while the remaining open panel stayed `data-behind`.
2. **Stack order never reaches the positioner (#3242):** `trackPanelStack` wrote the stack index (`--z-index`) only to the content element. Each positioner creates a sibling stacking context, so a z-index on the content can never reorder panels relative to each other — focusing a panel updated its state but it visually stayed behind.

## Fix

- The `closed` state now removes the panel from the stack on entry. The next panel is promoted to topmost immediately, and reopening re-adds the panel through the existing `open` entry action.
- The stack index is now also applied to the positioner element, and `getPositionerProps` consumes it via `zIndex: "var(--z-index)"`. The content mirror of `--z-index` is kept for userland styling.

## Tests

Added a two-panel `stacking` example page and two e2e regression tests (close-promotes-next and focus-raises-positioner, asserting `data-topmost` and computed positioner z-index). Both fail on `main` and pass with this change; verified locally with a rebuild in each direction.

### DIFF
--- a/.changeset/floating-panel-stack-close-zindex.md
+++ b/.changeset/floating-panel-stack-close-zindex.md
@@ -1,0 +1,11 @@
+---
+"@zag-js/floating-panel": patch
+---
+
+Fix panel stacking issues when multiple floating panels are open:
+
+- Closing a panel now removes it from the panel stack (previously only machine destroy did), so the next open panel is
+  correctly promoted to topmost.
+- The stack index is now applied to the positioner (`z-index` via `--z-index`), so focusing a panel actually brings it
+  above sibling panels. Previously the index was only mirrored on the content element, which cannot affect ordering
+  across sibling stacking contexts.
--- a/.changeset/floating-panel-stack-close-zindex.md
+++ b/.changeset/floating-panel-stack-close-zindex.md
@@ -2,10 +2,7 @@
 "@zag-js/floating-panel": patch
 ---
 
-Fix panel stacking issues when multiple floating panels are open:
+Fix panel stacking when multiple panels are open:
 
-- Closing a panel now removes it from the panel stack (previously only machine destroy did), so the next open panel is
-  correctly promoted to topmost.
-- The stack index is now applied to the positioner (`z-index` via `--z-index`), so focusing a panel actually brings it
-  above sibling panels. Previously the index was only mirrored on the content element, which cannot affect ordering
-  across sibling stacking contexts.
+- Closing a panel now removes it from the stack, so the next panel becomes topmost.
+- Stack order now applies to the positioner, so focusing a panel raises it above its siblings.
--- a/e2e/floating-panel.e2e.ts
+++ b/e2e/floating-panel.e2e.ts
@@ -203,4 +203,37 @@ test.describe("floating-panel", () => {
     await I.page.getByRole("button", { name: "API set size: 440x300" }).click()
     await I.seeContentHasSize(440, 300, 10)
   })
+
+  test("closing the topmost panel promotes the next panel to topmost", async ({ page }) => {
+    await I.goto("/floating-panel/stacking")
+
+    await page.getByTestId("trigger-A").click()
+    await page.getByTestId("trigger-B").click()
+
+    await expect(page.getByTestId("content-B")).toHaveAttribute("data-topmost", "")
+    await expect(page.getByTestId("content-A")).toHaveAttribute("data-behind", "")
+
+    await page.getByTestId("close-B").click()
+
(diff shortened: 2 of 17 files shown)


FrancoKaddour:
Thanks @segunadebayo — appreciated you jumping in on the branch; the bindable `stackIndex` + open-state `exit` is much cleaner than my ready-gated version. Great to see both stacking issues land together!

PR #3251, opened 2026-08-06
fix(dismissable): register escape handler synchronously when node is available

### LINKED ISSUE #3248: dismissable: Escape handler is deferred a frame for no reason, leaving dialogs undismissable while visible
## Description

`trackDismissableElement` defers its entire body by one `raf` when `defer: true`. That is correct for the interact-outside handler, but it also delays `trackEscapeKeydown`, which needs no deferral. The result is that a modal dialog is painted, focus-trapped and interactive-looking for at least one animation frame during which <kbd>Esc</kbd> does nothing.

Because the delay is pinned to `requestAnimationFrame`, it stretches with main-thread load rather than staying at one frame's worth of wall-clock.

Verified against `@zag-js/dismissable@1.43.0`, `@zag-js/dialog@1.43.0`, `@zag-js/interact-outside@1.43.0` (current latest at time of filing).

## Root cause

`dialog.machine.mjs` registers the layer with `defer: true`:

```js
trackDismissableElement(getContentEl, { type: "dialog", defer: true, ... })
```

`trackDismissableElement` then wraps *everything* in the deferred pass:

```js
function trackDismissableElement(nodeOrFn, options) {
  const { defer } = options;
  const func = defer ? raf : (v) => v();
  const cleanups = [];
  cleanups.push(
    func(() => {
      const node = isFunction(nodeOrFn) ? nodeOrFn() : nodeOrFn;
      cleanups.push(trackDismissableElementImpl(node, options)); // <- all of it
    })
  );
  ...
}
```

and `trackDismissableElementImpl` registers both handlers together:

```js
const cleanups = [
  pointerBlocking ? disablePointerEventsOutside(node, options.persistentElements) : void 0,
  trackEscapeKeydown(node, onEscapeKeyDown),                    // takes no `defer` — needs none
  trackInteractOutside(node, { ..., defer: options.defer })     // takes `defer` and applies it itself
];
```

Two things follow:

**1. The outer `raf` is not what protects against the opening click.** `trackInteractOutside` already implements the deferral internally — it re-wraps in `raf` at its own entry, wraps the `pointerdown` listener in a `setTimeout(…, 0)`, and defers again per-handler (skipping that last one on touch devices). So the interact-outside path is deferred whether or not the outer `raf` is there. The outer frame is redundant for it.

**2. Escape has nothing to defer against.** `trackEscapeKeydown` is a plain `document` keydown capture listener, guarded by `layerStack.isTopMost(node)` in `onEscapeKeyDown`. There is no opening gesture it could mistake for a dismissal, because a dialog cannot be opened with the Escape key. Its share of the window buys nothing.

The outer `raf` appears to exist only so the node can be resolved lazily — `getContentEl` may return `null` before the content commits.

## Impact

A dialog is visibly open and focus-trapped before <kbd>Esc</kbd> is wired up. Measured in a downstream Chakra UI v3 app (on 1.41.2, same code shape as 1.43.0), from the content getting `data-state="open"` to the layer being registered:

| CPU throttle | gap |
| --- | --- |
| none | ~16 ms |
| 6x | ~99 ms |
| 20x | ~360 ms |

An Escape sent at `data-state="open"` is ignored and the dialog stays open; the same key once focus has landed inside closes it. Under CPU contention this reliably swallowed a keypress in CI.

`layerStack.add(layer)` is also inside the deferred pass, so for that frame the layer is not on the stack at all.

## Suggested fix

Resolve the node synchronously and fall back to the deferred pass only when it isn't available yet, so `defer` continues to gate node resolution but no longer gates handler registration:

```js
function trackDismissableElement(nodeOrFn, options) {
  const { defer } = options;
  const cleanups = [];
  const setup = () => {
    const node = isFunction(nodeOrFn) ? nodeOrFn() : nodeOrFn;
    cleanups.push(trackDismissableElementImpl(node, options));
  };
  const node = isFunction(nodeOrFn) ? nodeOrFn() : nodeOrFn;
  if (!defer || node) setup();
  else cleanups.push(raf(setup));
  return () => cleanups.forEach((fn) => fn?.());
}
```

This keeps the interact-outside protection intact (it still defers itself), and preserves nesting order, since `layerStack.add(layer)` is the first statement in `trackDismissableElementImpl` and would now run before any nested layer could register.

If resolving the node eagerly is unwanted, the narrower alternative is to hoist `trackEscapeKeydown` out of the deferred pass alone, or to give `trackDismissableElementImpl` a separate flag so `defer` reaches only `trackInteractOutside`.

I'm happy to open a PR for whichever shape you'd prefer.

kirthi-b:
Fixes #3248

`trackDismissableElement` deferred its whole setup by one frame when `defer: true`, so a just-opened dialog was visible and focus trapped but ignored the Escape key until the next animation frame. Under CPU load that window stretches well past one frame.

The deferral only exists so the node can be resolved lazily, and the interact-outside logic already defers itself internally via its own `defer` option. This change resolves the node eagerly and registers handlers synchronously when it is available, keeping the deferred pass only as a fallback for nodes that have not rendered yet. The opening click still cannot dismiss the layer, and nested layers keep their stacking order since `layerStack.add` now runs before any later child registers.

Added unit tests covering immediate escape dismissal, the deferred fallback, topmost-only dismissal with nested layers, and cleanup.

### DIFF
--- a/.changeset/dismissable-escape-sync.md
+++ b/.changeset/dismissable-escape-sync.md
@@ -1,0 +1,7 @@
+---
+"@zag-js/dismissable": patch
+---
+
+Fix issue where the escape key handler and layer registration were deferred by a frame when `defer: true`, leaving a just-opened dialog undismissable via the escape key until the next animation frame.
+
+`trackDismissableElement` now resolves the node eagerly and registers handlers synchronously when the node is available. The deferred pass remains only as a fallback for nodes that have not rendered yet. Interact-outside behavior is unchanged, since it applies its own deferral internally.
--- a/packages/utilities/dismissable/src/dismissable-layer.ts
+++ b/packages/utilities/dismissable/src/dismissable-layer.ts
@@ -165,14 +165,25 @@ function trackDismissableElementImpl(node: MaybeElement, options: DismissableEle
 
 export function trackDismissableElement(nodeOrFn: NodeOrFn, options: DismissableElementOptions) {
   const { defer } = options
-  const func = defer ? raf : (v: any) => v()
   const cleanups: (VoidFunction | undefined)[] = []
-  cleanups.push(
-    func(() => {
-      const node = isFunction(nodeOrFn) ? nodeOrFn() : nodeOrFn
-      cleanups.push(trackDismissableElementImpl(node, options))
-    }),
-  )
+
+  // Resolve the node eagerly so the layer (and its escape keydown handler) is registered
+  // synchronously whenever possible. `defer` only needs to gate node resolution; the
+  // interact-outside logic applies its own deferral internally (via `options.defer`),
+  // so registering early cannot cause the opening click to dismiss the layer.
+  const node = isFunction(nodeOrFn) ? nodeOrFn() : nodeOrFn
+
+  if (!defer || node) {
+    cleanups.push(trackDismissableElementImpl(node, options))
+  } else {
+    cleanups.push(
+      raf(() => {
+        const deferredNode = isFunction(nodeOrFn) ? nodeOrFn() : nodeOrFn
+        cleanups.push(trackDismissableElementImpl(deferredNode, options))
+      }),
+    )
+  }
+
   return () => {
     cleanups.forEach((fn) => fn?.())
   }
--- a/packages/utilities/dismissable/tests/dismissable-layer.test.ts
+++ b/packages/utilities/dismissable/tests/dismissable-layer.test.ts
@@ -1,0 +1,133 @@
+// @vitest-environment jsdom
+
+import { afterEach, describe, expect, test, vi } from "vitest"
+import { trackDismissableElement } from "../src/dismissable-layer"
+import { layerStack } from "../src/layer-stack"
+
+function dispatchEscape(target: EventTarget = document) {
+  const event = new KeyboardEvent("keydown", { key: "Escape", bubbles: true, cancelable: true })
+  target.dispatchEvent(event)
+  return event
+}
+
+function nextFrame() {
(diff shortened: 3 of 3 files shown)


segunadebayo:
Thanks for putting this together, and for the clear write-up on the issue.

I've landed a fix in 9a9381d2d, so I'm closing this one — but I want to explain why it took a different shape, because your diagnosis of the mechanism was correct.

Resolving the node eagerly turns out not to fire in practice. Machine effects run *before* the framework commits its DOM, so `getContentEl` is always `null` at the moment `trackDismissableElement` is called. I instrumented both branches on `/dialog/basic` and got:

```
dialog open -> { sync: 0, raf: 1 }
```

The deferred pass is taken every time, so the timing is unchanged. Measuring the gap between the content entering the DOM and the keydown listener being attached confirmed it — ~33ms both with and without the eager path.

What does close it is a third step between the two: the node commits on the **microtask** following the call, which lands far sooner than the next frame. The landed fix escalates sync → microtask → rAF via a new `whenNode` helper in `@zag-js/dom-query`, used by both `trackDismissableElement` and `trackDismissableBranch`. That took the gap to ~0ms, and a real Escape keypress at 6x/20x CPU throttle from 4/5 to 5/5.

Your two supporting observations were both right and load-bearing for the fix: `trackInteractOutside` does defer itself (so registering early is safe), and `trackEscapeKeydown` needs no deferral at all. The one thing worth adding is that hoisting `trackEscapeKeydown` alone wouldn't have been enough either — `onEscapeKeyDown` bails on `layerStack.isTopMost(node)`, so the layer has to reach the stack in the same pass.

Really appreciate the report and the patch.

PR #3265, opened 2026-08-11
fix(focus-visible): use counter instead of boolean for pending event tracking

### LINKED ISSUE #10918: `Checkbox` and `Radio` (and maybe other components too) when used inside a `Dialog` or `Popover` (etc) component, they show a `focus-visible` ring
### `Checkbox` or `Radio`  component's expected behavior
- When I click, I want it to "focus" but not "focus-visible" and not show any outline.
- When I focus via keyboard navigation, I want it to be "focus-visible" and show the outline for accessibility.

Now,
This behavior works perfectly when using without any parent like `Dialog`, `Popover` etc.
But when I add the same checkbox or radio-group inside a Dialog, I see a `focus-visible` state on mouse click. And this we consider an inconsistent behavior.

Attaching a codesandbox for reference: https://codesandbox.io/p/sandbox/mkq2ph

What AI has to say:
> Mouse users just told you where their attention is by clicking; a ring is redundant visual noise, and it lingers after the interaction (looks especially odd on a checkbox you just ticked and moved on from). This is the majority view and what :focus-visible was invented for.

Please suggest if this a real bug or an intentional behavior. Also, is there a workaround for this? 
Thanks a lot Chakra team <3

Moreover my hacky fix was around below solution (but I am not satisfied with the slop hack). This is by AI:

# Root cause: @zag-js/focus-visible misclassifies pointer focus as virtual when a focus-trapped overlay re-homes focus mid-interaction
**Affected:** Checkbox, Switch, RadioGroup (any control whose zag machine reads isFocusVisible() on the hidden input's focus) when rendered inside a focus-trapped overlay (Popover / Menu / Dialog).

**Symptom:** Clicking such a control with the mouse paints the _focusVisible ring (data-focus-visible gets set), even though it's a pointer interaction. It only happens when focus moves to the control from another focused element inside the same overlay — so a standalone checkbox on a page is fine; a group of checkboxes in a popover is not.

**Mechanism (verified via CDP with trusted mouse events):**

When focus moves from control A to control B by clicking B, the event sequence is:

pointerdown  → target: B's control          modality = pointer
mousedown    → target: B's control
focusout     → A's hidden input
focusin      → the overlay's CONTENT container   ← focus-trap re-homes focus here first
...
focusout     → content container
focusin      → B's hidden input              modality = VIRTUAL   ← misclassified
The @zag-js/focus-visible global tracker gates its modality decision on an internal hasEventBeforeFocus flag. The intermediate focusin on the overlay's content container (injected by the dismissable-layer / focus-trap) runs handleFocusEvent, which consumes and resets hasEventBeforeFocus to false. So when the real, pointer-driven focus then lands on B's hidden input, handleFocusEvent takes the !hasEventBeforeFocus && !hasBlurredWindowRecently branch and sets currentModality = 'virtual'.

The checkbox/switch/radio machine's hidden-input onFocus reads isFocusVisible() synchronously at that moment — which is now true (virtual counts as focus-visible) — and sets data-focus-visible on the control. The recipe's focusVisibleRing / _focusVisible (&:is(:focus-visible, [data-focus-visible])) then renders the ring on a mouse click.

**In short: the focus-trap's intermediate programmatic focus on the overlay content clears hasEventBeforeFocus, so focus-visible loses the record of the immediately-preceding pointerdown and falls back to virtual for the subsequent genuine pointer focus.**

Where it seems fixable upstream (suggestions, not required):

> @zag-js/focus-visible: don't let a programmatic/container focus that isn't the interaction target reset hasEventBeforeFocus; or treat a focusin occurring within the same task as a recent pointerdown as pointer modality. Or: the focus machine reading isFocusVisible() only synchronously at onFocus is fragile against focus that arrives one tick after the pointer event — reconciling on the next microtask would also avoid it.
Reproduction shape: render ≥2 zag checkboxes (or switches/radios) inside a Popover, open it, then click one and click a different one with the mouse — the second gets data-focus-visible. Clicking the same control repeatedly does not trigger it (no focus move → no trap bounce). Keyboard Tab correctly sets data-focus-visible.

**Our workaround (for reference):** a small hook watches the control's own DOM — a pointerdown marks a pointer interaction; on the hidden input's subsequent focusin we queueMicrotask(() => setInteractionModality('pointer')), which re-asserts pointer after the machine's synchronous read and makes zag drop data-focus-visible in the same tick (no visible flash). keydown clears the flag so keyboard focus rings are untouched.

---

### Link to Reproduction

https://codesandbox.io/p/sandbox/mkq2ph?file=%2Fsrc%2FApp.js%3A165%2C23

### Steps to reproduce

2. Click on '...'
3. Scroll down to '...'

### Chakra UI Version

3.4

### Browser

Google Chrome

### Operating System

- [x] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

_No response_

waterWang:
## Problem

When a Checkbox, Radio, or Switch inside a focus-trapped overlay (Dialog, Popover, etc.) is clicked with the mouse, the `data-focus-visible` attribute is incorrectly set, causing a spurious focus ring to appear on a pointer interaction.

**Root cause:** The `hasEventBeforeFocus` boolean flag gets consumed by an intermediate programmatic focus event that the focus-trap / dismissable-layer fires on the overlay content container between the pointer event and the target element's focus. When the real target focus arrives, `hasEventBeforeFocus` is already `false`, so `handleFocusEvent` misclassifies the modality as `"virtual"` — which `isFocusVisible()` treats as focus-visible.

Fixes: chakra-ui/chakra-ui#10918

## Fix

Change `hasEventBeforeFocus` from a boolean to a counter (`pendingFocusEvents`). Each pointer/keyboard/virtual-click event increments the counter, and each focus event decrements it by one. Intermediate focus events from the focus-trap only consume one count, leaving the real target focus to still see a pending event.

## Testing

- Verified the logic matches the original contract for the non-focus-trap case: a single pointerdown to single focusin still works (counter goes 1 to 0).
- Window blur reset behaviour is preserved (counter set to 0).
- The `ignoreFocusEvent` fast-path is unchanged.

Closes chakra-ui/chakra-ui#10918

### DIFF
--- a/packages/utilities/focus-visible/src/index.ts
+++ b/packages/utilities/focus-visible/src/index.ts
@@ -62,7 +62,16 @@ interface GlobalListenerData {
 
 export let listenerMap = new Map<Window, GlobalListenerData>()
 
-let hasEventBeforeFocus = false
+/**
+ * Counter of pending keyboard/pointer events that haven't been matched by a focus event yet.
+ * A counter (not a boolean) is used because focus-trap / dismissable-layer can insert
+ * intermediate programmatic focus events (e.g. on the overlay content container) between
+ * the pointer event and the target element's focus. Each intermediate focus decrements
+ * the counter by one, so the final focus on the real target still sees an outstanding
+ * event and correctly keeps the "pointer" or "keyboard" modality — avoiding a spurious
+ * switch to "virtual" that would paint a focus-visible ring on a mouse click.
+ */
+let pendingFocusEvents = 0
 let hasBlurredWindowRecently = false
 
 /**
@@ -84,7 +93,7 @@ function triggerChangeHandlers(modality: Modality, e: HandlerEvent) {
 }
 
 function handleKeyboardEvent(e: KeyboardEvent) {
-  hasEventBeforeFocus = true
+  pendingFocusEvents++
   if (isValidKey(e)) {
     currentModality = "keyboard"
     triggerChangeHandlers("keyboard", e)
@@ -94,14 +103,14 @@ function handleKeyboardEvent(e: KeyboardEvent) {
 function handlePointerEvent(e: PointerEvent | MouseEvent) {
   currentModality = "pointer"
   if (e.type === "mousedown" || e.type === "pointerdown") {
-    hasEventBeforeFocus = true
+    pendingFocusEvents++
     triggerChangeHandlers("pointer", e)
   }
 }
 
 function handleClickEvent(e: MouseEvent) {
   if (isVirtualClick(e)) {
-    hasEventBeforeFocus = true
+    pendingFocusEvents++
     currentModality = "virtual"
   }
 }
@@ -123,12 +132,12 @@ function handleFocusEvent(e: FocusEvent) {
 
   // If a focus event occurs without a preceding keyboard or pointer event, switch to virtual modality.
   // This occurs, for example, when navigating a form with the next/previous buttons on iOS.
-  if (!hasEventBeforeFocus && !hasBlurredWindowRecently) {
+  if (pendingFocusEvents <= 0 && !hasBlurredWindowRecently) {
     currentModality = "virtual"
     triggerChangeHandlers("virtual", e)
   }
 
-  hasEventBeforeFocus = false
+  pendingFocusEvents = Math.max(0, pendingFocusEvents - 1)
   hasBlurredWindowRecently = false
 }
 
@@ -137,7 +146,7 @@ function handleWindowBlur() {
 
   // When the window is blurred, reset state. This is necessary when tabbing out of the window,
   // for example, since a subsequent focus event won't be fired.
-  hasEventBeforeFocus = false
+  pendingFocusEvents = 0
   hasBlurredWindowRecently = true
 }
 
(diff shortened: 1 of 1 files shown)


packages/utilities/focus-visible/src/index.ts
```diff
@@ -123,12 +132,12 @@ function handleFocusEvent(e: FocusEvent) {
 
   // If a focus event occurs without a preceding keyboard or pointer event, switch to virtual modality.
   // This occurs, for example, when navigating a form with the next/previous buttons on iOS.
-  if (!hasEventBeforeFocus && !hasBlurredWindowRecently) {
+  if (pendingFocusEvents <= 0 && !hasBlurredWindowRecently) {
```
sohamsshah, line 135: can update the comment to encompass new logic to add better context

sohamsshah, line 135: QUESTION: Can `pendingFocusEvents` go below 0?

segunadebayo:
Thanks for working on this. I have a better fix for the issue in commit `11926f072`, so I’m closing this PR.

PR #3267, opened 2026-08-11
fix(menu): prevent typeahead highlight from being overridden by scroll-induced pointermove

### LINKED ISSUE #3255: Menu's typeahead does not highlight an overflowed item if a mouse has hovered
# 🐛 Bug report

A menu's typeahead will bring the item into view if it is not visible. However, if I interact with a mouse cursor, and then trigger typeahead on an item that is off the screen, after the menu scroll, the item is deselected. If I try again without moving the mouse, it can be selected as expected.

## 💥 Steps to reproduce

1. Go to https://stackblitz.com/edit/kgnmn9q7
2. Click on the menu trigger to open the menu.
3. hover over the menu content.
4. Press <kbd>h</kbd> to bring the entry "Hey" into view.
5. It is selected and scrolled into view, but immediately it is deselected.

## 💻 Link to reproduction

https://stackblitz.com/edit/kgnmn9q7

https://github.com/user-attachments/assets/5f36a6be-9b91-4570-8c29-d3184a42b044

## 🧐 Expected behavior

The mouse hover should not affect the (typeahead + scroll into view) behavior.

## 🌍 System information

| Zag Version      | 1.43.0     |
| Browser          | Firefox 153/Chrom 146 |
| Operating System | GNU/Linux  |

waterWang:
## Problem

When the user hovers over the menu content with a mouse and then triggers typeahead (e.g. pressing h to find Hey), the matched item is highlighted and scrolled into view — but immediately deselected.

### Root cause

scrollToHighlightedItem scrolls the matched item into view. Because the menu items shift under the stationary cursor, the browser fires a pointermove event on a different item now under the cursor. That ITEM_POINTERMOVE event calls setHighlightedItem, overriding the typeahead highlight that was just set.

## Fix

Add an isTypingAhead guard to both ITEM_POINTERMOVE and ITEM_POINTERLEAVE so pointer events are ignored while typeahead search is active.

- Introduced a new isTypingAhead guard reading refs.typeaheadState.keysSoFar !== ""
- ITEM_POINTERMOVE: guard: and(not("isPointerRoutingLocked"), not("isTypingAhead"))
- ITEM_POINTERLEAVE: guard: and(not("isPointerRoutingLocked"), not("isTypingAhead"), not("isTriggerItem"))

The typeahead state has a built-in 350ms timeout (getByTypeahead), so normal pointer behavior resumes automatically once keysSoFar resets.

Fixes chakra-ui/zag#3255

### DIFF
--- a/packages/machines/menu/src/menu.machine.ts
+++ b/packages/machines/menu/src/menu.machine.ts
@@ -474,7 +474,7 @@ export const machine = createMachine<MenuSchema>({
         ],
         ITEM_POINTERMOVE: [
           {
-            guard: not("isPointerRoutingLocked"),
+            guard: and(not("isPointerRoutingLocked"), not("isTypingAhead")),
             actions: ["setHighlightedItem", "focusMenu", "closeSiblingMenus"],
           },
           {
@@ -482,7 +482,7 @@ export const machine = createMachine<MenuSchema>({
           },
         ],
         ITEM_POINTERLEAVE: {
-          guard: and(not("isPointerRoutingLocked"), not("isTriggerItem")),
+          guard: and(not("isPointerRoutingLocked"), not("isTypingAhead"), not("isTriggerItem")),
           actions: ["clearHighlightedItem"],
         },
         ITEM_CLICK: [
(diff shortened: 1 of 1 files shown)


segunadebayo:
Thanks for digging into this, and for the write-up — your root cause analysis was spot on: the scroll moves items under the stationary cursor and the resulting pointer event overrides the highlight that typeahead just set.

I've pushed a fix in c15e7ec36 / 4718e0b50 that takes a broader angle, so I'm closing this one.

Rather than gating on `isTypingAhead`, it gates on interaction modality (`getInteractionModality() !== "pointer"`). That covers a few cases the typeahead guard can't reach:

- **Arrow keys, Home and End.** These scroll far enough to move a different item under the cursor, but no typeahead is active, so `keysSoFar` is empty and the guard stays inert.
- **Select and combobox.** Both carried the same defect (they shared the stale `service.event.previous()?.type.includes("POINTER")` guard) and both reproduced on `End`.
- **Safari.** WebKit emits `pointermove` at an *unchanged* `clientX/clientY` when content scrolls under a resting cursor, which was flipping the global modality to `pointer` mid-keyboard-navigation. That's fixed in `focus-visible` so the modality signal is trustworthy for every machine.

Regression tests were added for menu, select and combobox — all seven fail without the fix.

Really appreciate the report and the patch.

PR #3278, opened 2026-08-14
feat(carousel): resume autoplay on visibility change

### LINKED ISSUE #3310: New opt-in prop for resume Carousel autoplay after document becomes visible again
# 🚀 Feature request

Add an opt-in `resumeOnVisibilityChange` option to the carousel's `autoplay` configuration. When autoplay is active, it should automatically resume after the document becomes visible again.

Carousel autoplay is paused when the browser tab or window becomes hidden. When the user returns to the page, autoplay remains paused and must be restarted manually.

This requires every application to implement its own `visibilitychange` handling or use workarounds, which makes carousel behavior inconsistent.

## ✅ Proposed solution or API

```tsx
const service = useMachine(carousel.machine, {
  autoplay: {
    delay: 4000,
    resumeOnVisibilityChange: true,
  },
})
```

Expected behavior:

- Autoplay pauses when the document becomes hidden.
- Autoplay resumes when the document becomes visible again.
- Autoplay resumes only if it was paused because of document visibility.
- A manually paused carousel remains paused.
- The default value is `false` for backwards compatibility.

The option should be available in the object form of the `autoplay` prop:

```ts
autoplay?: boolean | {
  delay: number
  resumeOnVisibilityChange?: boolean
}
```

## ↩️ Alternatives

- Implement `document.visibilitychange` handling in every consuming application.
- Always resume autoplay after the document becomes visible, which could override an intentional manual pause.
- Keep the current behavior and require users to restart autoplay manually.

## 📝 Additional Information

The implementation should use the Page Visibility API (`document.visibilityState` and `visibilitychange`).

Reference implementation: [PR #3278](https://github.com/chakra-ui/zag/pull/3278)

Current workaround implemented in our library: [Moduix Autoplay Gallery recipe](https://moduix.dev/recipes/autoplay-gallery)

E2E tests should cover both visibility-based resume and the case where manually paused autoplay remains paused.

Blinks44:
Closes #3310

## 📝 Description

Adds an opt-in `autoplay.resumeOnVisibilityChange` option to the carousel machine.

When the document becomes hidden, carousel autoplay is paused. When the document becomes visible again, autoplay remains paused.

## 🚀 New behavior

When `autoplay.resumeOnVisibilityChange` is enabled, autoplay resumes after the document becomes visible again only if it was paused by the document becoming hidden.

Explicit pauses through `api.pause()` or the autoplay trigger prevent automatic resumption.

The default behavior remains unchanged.

No.

## 📝 Additional Information

- No external dependencies added.
- Verified with:
  - `pnpm --filter @zag-js/carousel typecheck`
  - `pnpm --filter @zag-js/carousel lint`


Blinks44:
@segunadebayo Hello!
What do you think about this small new functionality in Carousel state machine?

Blinks44:
Forget about docs and examples, now added.

segunadebayo:
Thanks for the PR, and for the e2e test covering the manual-pause case.

I'm going to pass on this one. Two things decided it.

It moves `trackDocumentVisibility` from the `autoplay` state to the machine's root effects. Today that listener only exists while autoplay is running. After this change every carousel attaches a `visibilitychange` listener for its whole lifetime and sends `AUTOPLAY.PAUSE` on every tab switch, including carousels with autoplay off. That is a cost paid by everyone for a flag that defaults to `false`.

The `autoplayPausedByVisibility` flag also relies on the `autoplay` state's `AUTOPLAY.PAUSE` winning over the new root-level one. If that precedence ever shifts, the flag gets set and cleared in the same dispatch and the feature stops working silently, with the e2e test as the only thing guarding it.

The behavior is already reachable from the public API. Recipe in #3310.

PR #3286, opened 2026-08-17
feat(carousel): disable autoplay while manual scroll

Blinks44:
## 📝 Description

Fix carousel state synchronization after manual scrolling while autoplay is enabled.

When a user scrolls an autoplay carousel with touch or a horizontal mouse wheel/trackpad gesture, the carousel remains in the autoplay state. The internal page does not update after scroll snap completes, so indicators can remain on the previous slide and the next autoplay tick can navigate from stale state.

## 🚀 New behavior

Manual scrolling now moves the carousel from `autoplay` to the existing `userScroll` state. After scrolling settles, the carousel determines the closest page, updates indicators, and leaves autoplay paused.

No.

## 📝 Additional Information

- No public API or props were added.
- Added an e2e regression test for manual scrolling during autoplay.

Here is the video on my library page. Manual scrolling gets out of sync with autoplay.

https://github.com/user-attachments/assets/7f99c23f-1ca5-4a55-96f0-0cb219b1e337


Blinks44:
@segunadebayo Hi again :)

And here what do you think about fix this functionality or maybe add new prop?

Blinks44:
Nevermind, looks like `onWheel={api.pause}` and `onTouchStart={api.pause}` would be a better solution

PR #3291, opened 2026-08-19
fix(aria-hidden): use inert to suppress content below modals

### LINKED ISSUE #3203: Dialog backdrop passes through click events
3# 🐛 Bug report

Dialog backdrop can allow click-through when an underlying element explicitly sets `pointer-events: auto`.

When a Zag Dialog is opened, background content should not be interactive. However, the current pointer-blocking strategy appears to set `pointer-events: none` on `body`. If a descendant under the backdrop has `pointer-events: auto`, that element can still receive clicks even though it is visually covered by the backdrop.

Kind like #1720

## 💥 Steps to reproduce

1. Open Dialog
2. Try to hover or click the link element

## 💻 Link to reproduction

See: https://stackblitz.com/edit/3gfyhysi?file=src%2FDialog.tsx

## 🧐 Expected behavior

When a modal dialog is open, background content should not receive pointer events, regardless of whether descendants explicitly set `pointer-events: auto`.

The backdrop/positioner should reliably block interaction with all non-dialog content.

## 🧭 Possible Solution

1. Use an inert-based strategy for modal background suppression, such as `suppressOthers()` which already in zagjs (`suppressOthers` is a `inert`-frist prevented function), instead of relying only on `aria-hidden` and body-level `pointer-events: none`.

2.  A quick hack way is let your dialog backdrop element has `pointer-events: auto` to intercept and consume the click event.

## 🌍 System information

| Zag Version      | `@zag-js/dialog@1.41.2` |
| Browser          | Chrome |
| Operating System | Windows 11 |

Adebesin-Cell:
Closes #3203

## 📝 Description

modal content is suppressed with `inert` instead of `aria-hidden`, and `persistentElements` are kept out of the inert walk so they stay interactive.

content below a modal is hidden with `aria-hidden` and a body-level `pointer-events: none`. that value only reaches children by inheritance, so an element with `pointer-events: auto` breaks out and stays clickable through the modal.

## 🚀 New behavior

`ariaHidden` now uses `suppressOthers`, which sets `inert`. inert blocks clicks and focus regardless of `pointer-events`, and falls back to `aria-hidden` on browsers without inert support. applies to modal dialog, popover, and drawer.

`persistentElements` are kept interactive by setting `pointer-events: auto`, which inert would otherwise override and kill. they're now passed into `hideContentBelow` so they're excluded from the inert walk, matching how the dismissable layer already excludes them. dialog and popover; drawer has no `persistentElements` prop.

no. older browsers keep the `aria-hidden` path, and `persistentElements` keep working under a modal.

## 📝 Additional Information

changeset patches `@zag-js/aria-hidden`, `@zag-js/dialog`, `@zag-js/popover`, `@zag-js/drawer`.

### DIFF
--- a/.changeset/aria-hidden-inert-modal.md
+++ b/.changeset/aria-hidden-inert-modal.md
@@ -9,3 +9,6 @@ Fix background content receiving clicks through a modal overlay when an
 underlying element sets `pointer-events: auto`. Content below a modal is now
 made `inert` (falling back to `aria-hidden` where `inert` is unsupported), so
 it can't be clicked or focused regardless of its `pointer-events` value.
+
+`persistentElements` are excluded from the `inert` walk, so they stay
+interactive under a modal as before.
--- a/.changeset/aria-hidden-inert-modal.md
+++ b/.changeset/aria-hidden-inert-modal.md
@@ -1,0 +1,11 @@
+---
+"@zag-js/aria-hidden": patch
+"@zag-js/dialog": patch
+"@zag-js/popover": patch
+"@zag-js/drawer": patch
+---
+
+Fix background content receiving clicks through a modal overlay when an
+underlying element sets `pointer-events: auto`. Content below a modal is now
+made `inert` (falling back to `aria-hidden` where `inert` is unsupported), so
+it can't be clicked or focused regardless of its `pointer-events` value.
--- a/packages/machines/dialog/src/dialog.machine.ts
+++ b/packages/machines/dialog/src/dialog.machine.ts
@@ -197,7 +197,10 @@ export const machine = createMachine<DialogSchema>({
 
       hideContentBelow({ scope, prop }) {
         if (!prop("modal")) return
-        const getElements = () => [dom.getContentEl(scope)]
+        const getElements = () => [
+          dom.getContentEl(scope),
+          ...(prop("persistentElements")?.map((fn) => fn() as HTMLElement | null) ?? []),
+        ]
         return ariaHidden(getElements, { defer: true })
       },
     },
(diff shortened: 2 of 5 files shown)


FrancoKaddour (commented):
Really like this move to `inert` — it's the correct primitive for "un-interactive subtree," and it fixes the `pointer-events: auto` leak at the root instead of patching each case. The `persistentElements` exclusion from the walk is a nice touch too.

Two things worth considering, both around the edges rather than the core:

**1. The `aria-hidden` fallback path looks untested.** The changeset promises "falling back to `aria-hidden` where `inert` is unsupported," but since the test monkeypatches `HTMLElement.prototype.inert` to force the inert branch, the fallback branch of `suppressOthers` never gets exercised. Given Safari <15.4 and some embedded webviews still lack `inert`, it might be worth a sibling test that deletes the `inert` property (or stubs `'inert' in HTMLElement.prototype` to `false`) and asserts the elements get `aria-hidden="true"` + the pointer-events guard instead. Otherwise a regression in that path would ship silently.

**2. Focus behavior when `inert` wraps the active element.** Marking an ancestor `inert` will blur whatever was focused inside it and reset focus to `<body>`. For a modal that's usually fine (focus moves into the dialog), but if a focused node ends up inside the inert region — e.g. a `persistentElement` that *isn't* excluded, or the trigger in the popover case — the user could see an unexpected focus jump. Might be worth confirming the active trigger / persistent elements are always part of the non-inert set. Not a blocker, just want to make sure the focus contract holds.

Approach LGTM either way — these are hardening questions, not objections. 👍

PR #3303, opened 2026-08-29
feat(core): add watchEffect for effects that re-run on dep change

### LINKED ISSUE #3302: Make dialog open-state props (trapFocus, preventScroll, modal) reactive, and mark non-reactive props in the types
# 🚀 Feature request

Make the dialog machine's open-state props (`trapFocus`, `preventScroll`, `modal`) reactive, so toggling them on an already-open dialog takes effect immediately. Today they are read once on entry to the `open` state, so the only way to change them is a `key` remount.

This follows on from [discussion #10128](https://github.com/chakra-ui/chakra-ui/discussions/10128), where `trapFocus` and `modal` were decoupled so a non-modal dialog can still trap focus. That decoupling is what makes reactivity worth having: the dialogs that need `modal={false}` with `trapFocus={true}` are usually the ones that need to *change* that behaviour partway through their lifetime, without closing.

**Our use case: a media player in a dialog that can be maximized or minimized.**

While maximized, the player should trap focus and lock body scroll. When the user minimizes it, it becomes a non-blocking picture-in-picture — focus and scroll must be released so the rest of the page stays usable. The dialog never closes across that transition; only its blocking behaviour changes.

The `key` remount workaround doesn't work here. Remounting the dialog remounts the player inside it, which tears down and recreates the underlying media element. The result is an audible/visible break in playback — an abrupt jitter or pause in the audio and video every time the user minimizes or maximizes. For a learning platform where people listen to podcasts and watch videos while browsing, losing playback continuity on every minimize is not an acceptable trade.

So the workaround that exists for `modal` is specifically unusable for the case that motivates the feature: any dialog with continuous media, a running timer, a live connection, or any other state that cannot survive a remount.

Concretely, in the current machine (`@zag-js/dialog@1.41.2`), the effects are registered on state entry:

```js
states: {
  open: {
    entry: ["checkRenderedElements", "setInitialFocus"],
    effects: ["trackDismissableElement", "trapFocus", "preventScroll", "hideContentBelow"],
```

and each one gates on its prop at setup time:

```js
trapFocus({ scope, prop, context }) {
  if (!prop("trapFocus")) return
  ...
}

preventScroll({ scope, prop }) {
  if (!prop("preventScroll")) return
  return preventBodyScroll(scope.getDoc())
}

hideContentBelow({ scope, prop }) {
  if (!prop("modal")) return
  ...
}
```

The machine's only watcher is on `open`:

```js
watch({ track, prop }) {
  track([() => prop("open")], () => {
    action(["toggleVisibility"])
  })
}
```

Because nothing tracks `trapFocus`, `preventScroll`, or `modal`, an effect that returned early at setup is never re-run, and an active one is never torn down, for as long as the dialog stays open.

This matches what was acknowledged in #10128 for `modal` — that the open-state effects read `modal` when they are set up on dialog entry and don't re-subscribe when it changes. This request is to address the same limitation for the whole set, since `trapFocus` and `preventScroll` are the ones most useful to toggle live.

**Steps to reproduce:**

1. Render a dialog with `modal={false}` and `trapFocus={true}`.
2. While it is open, flip `trapFocus` to `false`.
3. Focus remains trapped. Same for `preventScroll` (body stays locked) and `modal`, in both directions.

## ✅ Proposed solution or API

No API change — the existing props simply become reactive. This should just work:

```tsx
const isBlocking = !isMinimized

<Dialog.Root
  open
  modal={false}
  trapFocus={isBlocking}
  preventScroll={isBlocking}
>
  <MediaPlayer />   {/* keeps playing across the transition */}
</Dialog.Root>
```

Implementation-wise, tracking the props that gate open-state effects so they re-subscribe on change, alongside the existing `open` watcher:

```js
watch({ track, prop }) {
  track([() => prop("open")], () => {
    action(["toggleVisibility"])
  })
  track([() => prop("trapFocus")], () => { /* re-run the trapFocus effect */ })
  track([() => prop("preventScroll")], () => { /* re-run the preventScroll effect */ })
  track([() => prop("modal")], () => { /* re-run modal-gated effects */ })
}
```

The mechanism is yours to choose — the goal is that these effects are re-evaluated when their gating prop changes while the dialog is open, instead of only on entry.

### Making reactivity visible in the types

Separately, and useful regardless of whether the above lands: there is currently no way to tell from the types or the docs which props are safe to change on a live machine. `open` is reactive, `trapFocus` is not, and both are plain booleans on the same interface. We only discovered the difference by reading `dialog.machine.mjs`.

If some props are going to stay non-reactive — here, or in other machines — it would help a lot to say so in the type itself, so it shows up on hover in the editor rather than being found at runtime:

```ts
export interface DialogProps {
  /**
   * Whether the dialog is open.
   */
  open?: boolean
  /**
   * Whether to trap focus inside the dialog.
   *
   * @nonReactive Read once when the dialog opens. Changing this while the
   * dialog is open has no effect until it is remounted or reopened.
   */
  trapFocus?: boolean
}
```

A JSDoc tag like `@nonReactive` costs nothing at runtime and is greppable across machines. A branded type (`NonReactive<boolean>`) would be stronger, though probably more disruptive than it's worth. Either way, the win is that the constraint is discoverable at the call site.

## ↩️ Alternatives

1. **`key` remount** (the workaround from #10128) — breaks playback continuity, which is the problem we're trying to solve.
2. **Hand-rolled focus trap and scroll lock** via `react-focus-on`, with `trapFocus={false}` / `preventScroll={false}` on the dialog. This is what we do today; it works because the wrapper's `enabled` prop is reactive, but it means running a second focus-management system alongside the dialog's own, which interacts badly with nested dialogs.
3. **Closing and reopening the dialog** on each transition — same remount problem, and it loses the open/close animation semantics.

## 📝 Additional Information
- Discussion this follows from: https://github.com/chakra-ui/chakra-ui/discussions/10128
- Version inspected: `@zag-js/dialog@1.41.2` (same behaviour in `1.21.7`)
- Code references above are from `dist/dialog.machine.mjs` in that version.

segunadebayo:
Closes #3302.

`trapFocus`, `preventScroll` and `modal` were read once when a dialog opened, so changing them required a remount, which destroys whatever the dialog contains.

### `watchEffect(deps, setup)`

Re-runs a gated effect when its deps change, without re-entering the state. Mirrors `track` in `watch`: same shape, same `DepFn` constraint, one runs an action and the other an effect.

```ts
trapFocus({ prop, watchEffect }) {
  return watchEffect([() => prop("trapFocus")], () => {
    if (!prop("trapFocus")) return
    return trapFocus(/* ... */)
  })
}
```

Only that effect restarts. Sibling effects and entry actions are untouched, unlike `reenter: true`. Applied to 9 effects across dialog, drawer and popover. Demo at `/dialog/reactive-props`.

**Needs review:** releasing the focus trap no longer returns focus to the trigger. Focus is restored only when the overlay actually closes.

### Also fixed

Both pre-existing and unrelated to the feature:

- Preact `Portal` imported hooks from `preact/compat`, which gets its own hooks registry under a bundler. Nothing portalled rendered at all. `useSyncExternalStore` had the same exposure.
- Preact's test suite could not run, since its tsconfig did not override the root's `jsx: "preserve"`. Fixing it resurrected 28 tests.

### Testing

616 unit tests across core, machines and all five adapters. 52 E2E across dialog, popover and drawer.

### Not included

`Portal` gives each child its own `createPortal`, so a conditionally rendered backdrop mounts after the positioner and covers the content. A one-line fix works and passes 137 E2E, but it changes shared behaviour. The example keeps its backdrop mounted instead.


segunadebayo:
Superseded by #3304, which targets `v2`.

`v2` already contains all of `main` plus 213 commits and is the active release line, and it carried the same bugs: the position-based `getNestedLayers`, and `z-index: var(--layer-index)` on both backdrop and positioner. Landing here would have meant porting the work across anyway.

The port is not a straight cherry-pick: `v2` had independently rewritten `layer-stack.ts` (adding `LayerSnapshot`/`onLayerChange`) and `pointer-event-outside.ts` (replacing the per-layer MutationObserver with a `data-inert` model), so the dismissable changes were redone against those shapes.

Issue #2241, opened 2025-02-17
Popovers close instantly when it is opened while also having another Popover open


blenderskool:
# 🐛 Bug report

_Duplicated from https://github.com/chakra-ui/ark/issues/3287_

When I have multiple Popover like components (Popover, Menu, Select, etc.) and one of them is open, if I try to open another Popover like component, it sometimes instantly closes as soon as it opens. I would expect it to open and only the earlier Popover component to close.

This issue is also visible on Ark UI, Chakra UI documentation pages.

## 💥 Steps to reproduce

1. Go to https://codesandbox.io/p/sandbox/ynp3yv
2. Single click on "Popover 1". Popover 1 should open.
3. Single click on "Popover 2". Popover 2 should open and then instantly close. (It might happen after multiple tries)

## 💻 Link to reproduction

CodeSandbox reproduction: https://codesandbox.io/p/sandbox/ynp3yv

## 🧐 Expected behavior

Content of Popover like component that is opened to remain open and not get closed when other Popover like element closes.

## 🧭 Possible Solution

I was trying to create a PR with a fix and I figured out that when the second popover element that should stay open closes instantly, it is because of the elements being dismissed by layerStack. This is the sequence of events that I was able to notice:
1. Second popover opens.
2. First popover closes because of onInteractOutside / onFocusOutside.
3. Closing of first popover results in nested layers of layerStack to also get removed.
4. Because second popover opened when first popover was open, it gets considered as a nested popover and gets closed by layerStack as its _parent element_ (first popover) closed.

Two fixes I could identify:
1. Deferring opening of a popover:
    ```ts
        onClick(event) {
          if (event.defaultPrevented) return
          setTimeout(() => send("TOGGLE"), 0)
        },
    ```
2. Disabling defer on trackDismissableElement in popover:
    ```ts
        return trackDismissableElement(getContentEl, {
            defer: true,
           // ...
       })
    ```
I don't know if any of these are the right solutions to this, as it is something that affects other Popover like components (Menu, Select, etc.) too and there might be some other underlying issue.

## 🌍 System information

| Software         | Version(s) |
| ---------------- | ---------- |
| Zag Version      | 0.82.1           |
| Browser          | Chrome           |
| Operating System | macOS           |

## 📝 Additional information

https://github.com/user-attachments/assets/5952507a-934d-46b3-b652-b86f6f8b3a45

segunadebayo:
I just pushed a fix for this. We'll release an update shortly.

[...]

PR #3322, opened 2026-09-07
fix(dismissable): stop a layer from dismissing the layers beside it

itkrt2y:
## 📝 Description

Opening a select, popover or menu while another one is open closes the first (as expected) and then immediately closes the one just opened. `layerStack` inferred nesting from stack position, so two layers opened in sequence were indistinguishable from a dialog opened from inside a dialog. This makes nesting explicit: a layer resolves the layer it is nested in when it is added, and removal dismisses only that layer's own subtree.

`getNestedLayers(node)` returned every layer above `node`, while `isInNestedLayer` layered DOM containment on top of that. Two layers opened side by side satisfy the first notion and not the second, and nothing else told them apart.

Whether the bug surfaces depends on a one-frame window rather than on which components are involved. `trackInteractOutside` defers its outside check to the next animation frame, a select opens on `click`, and a deferred layer reaches the stack on a microtask when its node has already committed — 9a9381d dropped the frame it used to wait for. The layer being opened therefore reaches the stack before the deferred check runs, whenever the press is shorter than one frame. A leisurely mouse press on an idle page stays clear of that window; a fast click, a programmatic click, or a busy main thread that delays the frame does not. On touch it is not a race at all: the outside check runs from a `click` listener on `document`, so event propagation guarantees the newly opened layer is registered first.

## 🚀 New behavior

A layer resolves its parent once, when it is added: the layer whose node contains it, or — nested content is usually portalled — the layer containing one of its `exclude` containers, which machines fill with the trigger the layer is opened from. `persistentElements` is deliberately kept out of that lookup, since it says what must not dismiss a layer, not where the layer came from. A layer opened from outside every open layer gets no parent, and `remove` walks only the subtree beneath the removed layer.

Removal has to keep that parentage intact. Dismissing a child can synchronously remove other layers, so `remove` re-reads its own index after the cascade instead of reusing the one it started with. A child that outlives the layer it hangs under — its dismissal was prevented, or it is still animating out — is handed to that layer's own parent rather than left pointing at a node no longer on the stack, so an ancestor still dismisses it.

`data-nested`, `data-has-nested` and `isInNestedLayer` all read the same parentage now, so the three agree on what nesting means.

No — no public API changes. One observable behavior change is worth calling out: an interaction inside a layer that is not nested in the receiving layer now reaches that layer's `onInteractOutside`, where previously any layer higher on the stack counted as inside and suppressed it.

## 📝 Additional Information

Reproduction: open `/select/multiple-controlled` in the `next-ts` example, open the first select, then click the second.

### Before

https://github.com/user-attachments/assets/a1be5f96-c5fd-4d41-a226-d08bdd2b876b

### After

https://github.com/user-attachments/assets/ed31c949-dd8e-418f-a659-1d1590840fe1

Left uncovered: a layer with no trigger rendered and no containing layer falls back to the top-most layer, so it can still take a sibling down with it — `navigation-menu` passes no `exclude` at all and always lands there. `exclude` also usually holds more than one element: every machine but `color-picker` and `navigation-menu` passes the trigger together with a clear trigger or the input, or every trigger sharing the owner id. Nesting resolves to the top-most layer containing any of them, so a layer with one of those elements inside another layer and the rest outside resolves to the inner layer whichever element the interaction came from. Resolving that needs the element that actually fired, which is a larger change than this one.

The same symptom was reported in #2241 and addressed then by removing a deferred frame from the teardown side of the race ( 7519355 ). Resolving nesting rather than ordering it removes the dependency on which side of that race wins.

<details><summary><h3>Greptile Summary</h3></summary>

This PR replaces stack-position-based dismissable-layer nesting with explicit parent relationships inferred from DOM containment or trigger containers.
- Limits cascading dismissal to a removed layer’s own descendant subtree.
- Reparents surviving children when an intermediate layer is removed.
- Aligns outside-interaction checks and nested-layer metadata with the same ancestry model.
- Adds unit and end-to-end coverage for sibling layers, portalled children, subtree removal, and re-registration.
</details>

<h3>Confidence Score: 5/5</h3>

The PR appears safe to merge, with the remaining parent-inference limitations explicitly documented and no additional actionable defect established.

Explicit ancestry consistently drives subtree removal, outside-interaction suppression, and nesting metadata, while the added tests cover the principal sibling-layer regression and important teardown cases.

<details><summary><h3>Important Files Changed</h3></summary>

| Filename | Overview |
|----------|----------|
| packages/utilities/dismissable/src/layer-stack.ts | Introduces explicit layer ancestry, subtree-based dismissal, survivor reparenting, and ancestry-derived nesting metadata; no publishable defect was established. |
| packages/utilities/dismissable/src/dismissable-layer.ts | Separates trigger containers from persistent exclusions and supplies trigger candidates for parent inference. |
| packages/utilities/dismissable/tests/layer-stack.test.ts | Adds focused coverage for sibling independence, portalled nesting, transitive dismissal, survivor handoff, and duplicate registration. |
| packages/utilities/dismissable/tests/dismissable-layer.test.ts | Adds integration-level coverage for the deferred outside-interaction race and persistent-element semantics. |
| e2e/select-multiple-controlled.e2e.ts | Verifies that opening a second controlled select closes the first without immediately closing itself. |

</details>

<details open><summary><h3>Sequence Diagram</h3></summary>

```mermaid
sequenceDiagram
  participant Trigger as Sibling trigger
  participant Existing as Existing layer
  participant Stack as Layer stack
  participant New as Newly opened layer
  Trigger->>Existing: Outside interaction begins
  Trigger->>Stack: Register new layer
  Stack->>Stack: Resolve parent from containment/trigger
  Stack-->>New: No parent (sibling)
  Existing->>Stack: Remove existing layer
  Stack->>Stack: Collect only Existing descendants
  Stack-->>New: Preserve unrelated sibling
```
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(dismissable): stop a layer from dism..."](https://github.com/chakra-ui/zag/commit/43674a2139b386cbbc099e4abdc1665bea8e5327) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=61126636)</sub>


FrancoKaddour (commented):
Nice approach — resolving the parent explicitly at add time (node containment → trigger containment → stack-order fallback) is a real improvement over inferring nesting from stack position, and the details show care: the orphan handover to `layer.parent` on removal, re-reading the index after dismissing children synchronously, and `clear()` looping now that removing `layers[0]` no longer cascades through everything above it. The changeset honestly documenting the `onInteractOutside` behavior change is appreciated too.

One heads-up, verified locally: this needs a one-line update when it rebases onto `main`. de9aeaaf8 (landed today, from #3323) makes `defer: true` always defer registration to the microtask — it no longer runs synchronously when the node already exists. Your new test asserts synchronously:

```ts
const cleanup = trackDismissableElement(node, { defer: true, exclude: [trigger], onDismiss })
expect(layerStack.isTopMost(node)).toBe(true) // fails post-merge — the layer lands on the microtask now
```

Ran the suite three ways: this branch as-is (28/28 ✅), merged with `main` (that assertion fails, plus two pre-existing failures on `main` itself that #3337 addresses), and merged with an `await Promise.resolve()` added before the assertion (28/28 ✅). The rest of the test survives always-defer untouched — the microtask add still precedes the rAF outside-check, so parent resolution keeps seeing the pre-dismissal stack, which is exactly what your sibling logic needs.

One question, non-blocking: `triggerElements` reuses `excludeContainers`, which overloads `exclude` ("don't dismiss on this") as "opened from here". You clearly saw the hazard — the persistent-elements test pins that those don't create nesting. Is there a machine that passes something unrelated in `exclude` that lives inside another open layer (and would now nest into it)? If that's a real risk, a dedicated option might be safer long-term — but as-is the trigger heuristic covers the reported bug well. LGTM once the rebase lands.

itkrt2y:
Thank you for catching the `when-node` change before it landed. Rebased onto `main` and added the `await Promise.resolve()` you suggested, folded into the same commit ( 14abfea31 ). The suite is 26/28 here, and the two failures are exactly the pre-existing ones on `main` that #3337 addresses — both test bodies are byte-identical to the ones already on `main`, so nothing in this PR touches them. The `select-multiple-controlled` e2e still passes, and still fails when I swap only `packages/utilities/dismissable/src` back to `main`, so the race window does survive always-defer as you expected.

> Is there a machine that passes something unrelated in `exclude` that lives inside another open layer (and would now nest into it)?

No. Every element a machine puts in `exclude` is resolved within that component instance, so none of them can be something unrelated.

The layer-registering callers and what they pass:

- `select`, `cascade-select` — trigger, clear trigger
- `combobox` — input, trigger, clear trigger
- `date-picker` — the inputs, trigger, clear trigger
- `color-picker` — trigger
- `dialog`, `drawer`, `hover-card`, `menu`, `popover` — trigger plus every trigger sharing the owner id
- `navigation-menu` — nothing, the fallback case called out in the commit message

`toast` and `tour` use `trackDismissableBranch`, which registers a branch rather than a layer and takes no `exclude`.

Each of those is scoped to the instance: the single elements come from `getById` on that instance's own id, `date-picker`'s inputs are queried under its own control element, and `getTriggerEls` filters on `getByOwnerId(scope.id)`.

The one place an arbitrary element could enter is `persistentElements`, which `dialog` and `popover` pass straight through from the user's prop, and parent resolution deliberately skips it: the layer carries `triggerElements: getExcludeContainers`, not `getExcludedElements`. `persistentElements` says what must not dismiss a layer, not where it was opened from, and the test you mentioned pins that.

What can still misfire is not an unrelated element but the multi-element case, and it reaches further than the "Not covered" note originally said, so I have widened that note in the commit message and the PR description. `findParentNode` returns the top-most layer containing any of the `exclude` elements, and every machine but `color-picker` and `navigation-menu` passes more than one — the trigger together with a clear trigger or the input, or every trigger sharing the owner id. Nothing forces those to render in the same place, so an app placing one of them inside another open layer and the rest outside resolves to that layer whichever element the interaction actually came from.

That last part is also why I do not think a dedicated option fixes it on its own: splitting "opened from here" out of `exclude` would remove the overloading, but the set would still hold several triggers and the ambiguity would remain. What resolves it is knowing which element actually fired, which is the larger change.

FrancoKaddour:
Perfect — folding it into the same commit keeps the history clean, and thanks for re-running the three-way check. Once #3337 lands, the two `main` failures clear and this goes fully green. Nice work on the sibling-layer model. 👍

itkrt2y:
Rebased onto `main` ( def7dcb58 ). #3337 has landed since my last comment, so the two pre-existing failures I mentioned are gone — the dismissable suite is 28/28 here, and the `select-multiple-controlled` e2e still passes.

PR #3338, opened 2026-09-11
fix(scripts): derive package names with basename for Windows

dajiaohuang:
## 📝 Description

`scripts/` feeds paths from three sources into string splitting: `find-packages` (platform-native absolute paths), `fast-glob` (always forward slashes) and `resolve()` (platform-native). Splitting by the wrong separator silently returns the whole path or `undefined` instead of the package name. This replaces the manual splitting with `basename`/`dirname` in the four affected scripts.

On Windows, where `sep` is `\`:

- `pnpm check-type-exports` reports every index export of all 51 machines as missing — 699 false failures — and exits 1.
- `pnpm document-types` (run by `pnpm build` through `postbuild`) rewrites `packages/docs/data/api.json` with 51 keys that are absolute file paths and empty `api` / `context` objects. It exits 0, so the corruption is silent, and because `document-attr` runs next it also leaves the file on disk.
- `pnpm document-attr` throws `TypeError: Cannot read properties of undefined (reading 'toLowerCase')` and exits 1, so `pnpm build` fails outright.
- `pnpm document-css-vars` collapses every connect/style variable under a single `undefined` key and loses the popper lookup, dropping 15 of 31 components and all of the positioner variables.

On macOS and Linux the paths are already `/`-separated, so none of this is visible.

## 🚀 New behavior

All four scripts resolve the package name on every platform. `check-type-exports` prints `All 51 machines have correct type exports! ✨` and exits 0, and `document-types`, `document-attr` and `document-css-vars` all exit 0 and regenerate the checked-in files.

No.

## 📝 Additional Information

Verified on Windows 10 with Node 24.14.1 and pnpm 11.22.0 against `a40f3f34a`:

| script | before | after |
| --- | --- | --- |
| `check-type-exports` | 699 failures, exit 1 | exit 0 |
| `document-types` | `api.json` rewritten with absolute-path keys | matches the committed file |
| `document-attr` | `TypeError`, exit 1 | exit 0, `data-attr.json` byte-identical to committed |
| `document-css-vars` | 16 of 31 components, no popper vars | 31 components, matches the committed file |

Every regenerated file was compared against the committed blob: `data-attr.json` is byte-identical; `api.json` differs only by `\n` → `\r\n` inside JSDoc strings, where all 157 differences are explained by this checkout's `core.autocrlf=true`; `css-vars.json` is JSON-equal with only filesystem-dependent key ordering. ESLint and Prettier are clean.

No changeset: only `scripts/` is touched and the content of the generated files is unchanged, so there is no user-facing change to describe. That follows `.claude/docs/changelog-guide.md` ("Skip internal refactors") and the precedent of `921419e59`, which changed `scripts/css-vars.ts` and `scripts/data-attr.ts` without one.

### DIFF
--- a/scripts/check-type-exports.ts
+++ b/scripts/check-type-exports.ts
@@ -1,4 +1,4 @@
-import { join } from "path"
+import { basename, join } from "path"
 import {
   EnumDeclaration,
   InterfaceDeclaration,
@@ -190,7 +190,7 @@ async function main() {
   let totalIssues = 0
 
   for (const { dir } of machines) {
-    const machineName = dir.split("/").pop()!
+    const machineName = basename(dir)
     const result = checkMachine(project, dir, machineName)
 
     if (result.issues.length > 0) {
--- a/scripts/css-vars.ts
+++ b/scripts/css-vars.ts
@@ -1,4 +1,4 @@
-import { resolve } from "node:path"
+import { basename, dirname, resolve } from "node:path"
 import { writeFileSync, readFileSync } from "node:fs"
 import { glob } from "fast-glob"
 import { Project, Node, SyntaxKind } from "ts-morph"
@@ -223,7 +223,7 @@ function extractCSSVariablesFromConnect(filePath: string, project: Project): CSS
   if (!sourceFile) return []
 
   const variables: CSSVariable[] = []
-  const component = filePath.split("/").slice(-3, -2)[0] // Extract component name from path
+  const component = basename(dirname(dirname(filePath)))
 
   sourceFile.forEachDescendant((node) => {
     if (Node.isPropertyAssignment(node) && node.getName() === "style") {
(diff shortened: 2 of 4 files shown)


dajiaohuang:
No changeset is included here on purpose: this PR only touches `scripts/`, and the content of the generated files it writes is unchanged, so there is no user-facing change to describe. `.claude/docs/changelog-guide.md` says to skip internal refactors, and the closest precedent — `921419e59 chore: update scripts`, which changed `scripts/css-vars.ts` and `scripts/data-attr.ts` — added no changeset, while changesets in this repo accompany edits to published package source (`a372b9559`, `5d8f638ac`, `fe6edba68`). Happy to add one if you would rather have `.cursor/rules/development.mdc` apply to tooling-only changes too.

PR #3352, opened 2026-09-18
fix(number-input): don't overwrite a keystroke that lands before the input sync

### LINKED ISSUE #2680: number-input: displays incorrect values in tests
# 🐛 Bug report

Number Input displays incorrect values when typing - the component has a non-deterministic bug where typed characters appear in wrong order during testing.

## 💥 Steps to reproduce
- Clone and run the test:
```sh
git clone https://github.com/sairus2k/zagjs-number-input-issue
cd zagjs-number-input-issue
npm install
npm test
```

- The bug occurs inconsistently, with a frequency ranging from 5% for short numbers (three digits) to 40% for long numbers (ten digits).

## 💻 Link to reproduction

GitHub reproduction: https://github.com/sairus2k/zagjs-number-input-issue

## 🧐 Expected behavior

Input should display characters in the exact order they were typed — "123456789" should display as "123456789", not scrambled

## 🧭 Possible Solution

This appears to be a race condition in zagjs's input handling when processing rapid synthetic events from testing-library. Since:

- Native inputs work correctly with the same test setup
- Manual browser testing doesn't reproduce the issue
- The same behaviour when using jsdom instead of the Vitest browser mode.

## 🌍 System information

| Zag Version      | 1.22.1           |
| Browser          | Firefox 143           |
| Operating System | macOS 15.6.1           |
| Test Framework | Vitest with browser testing |
| Testing Library | @testing-library/user-event |

## 📝 Additional information

**Impact**: Critical for testing - automated tests fail intermittently, making CI/CD unreliable.

**Actual behavior observed**:

- "123" sometimes becomes "132" (low frequency ~5%)
- "1.5" sometimes becomes "15." (low frequency ~5%)
- "123456789" sometimes becomes scrambled (high frequency ~40%)
- Longer inputs = higher failure rate

**Important notes**:

- Bug ONLY occurs when using @testing-library/user-event
- Manual testing in browser works correctly
- Native HTML <input type="number"> works correctly with identical test setup
- This confirms the issue is specific to how zagjs processes synthetic events from testing-library

**Workaround**: use `paste` userEvent method instead of `type`

strongpauly:
Closes #2680

## 📝 Description

`syncInputElement` queues a `raf` that writes back the text and cursor recorded when the event arrived:

```ts
syncInputElement({ context, event, computed, scope }) {
  const value = event.type.endsWith("CHANGE") ? context.get("value") : computed("formattedValue")
  const inputEl = dom.getInputEl(scope)
  const sel = event.selection ?? recordCursor(inputEl, scope)
  raf(() => {
    setElementValue(inputEl, value)     // text from before the event
    restoreCursor(inputEl, sel, scope)  // cursor from before the event
  })
}
```

Both are written unconditionally a frame later. A keystroke that lands inside that window gets
overwritten: the input reverts to the previous text and cursor, the following character is inserted
at the stale offset, and the reverted character resurfaces behind it.

This is #2680 — closed with the suggestion to add a delay in tests, which works but leaves the
underlying race in place, and (as the reporter noted there) needs retuning per machine.

Typing faster than one character per animation frame drops or transposes characters. Instrumenting
the input across a run that reproduces it:

```
typed "5": ""@0  -> "5"@1
typed ".": "5"@1 -> "5"@1     <- the "." is gone again
typed "5": "5"@1 -> "5."@2     <- it comes back, and the new "5" lands at index 1
settled  "55."   ->  blurred "55"
```

So `1.5` becomes `15.`, and with `maximumFractionDigits: 0` that commits as `55`. Longer entries are
hit harder because there are more windows to land in, which matches the rates reported in #2680
(~5% at three digits, ~40% at ten).

It is not specific to a test runner. I hit it first through `@storybook/test-runner` in a real
Chromium, but it reproduces on a bare `NumberInput.Root` with nothing else rendering it and no
cursor fed back by the harness. Under load it is reachable by hand: on a machine where a frame takes
100 ms+, an ordinary typist can put two keys in one frame.

## 🚀 New behavior

The queued write is skipped when the input has been typed into since it was queued — that keystroke
has queued a sync of its own, so nothing is lost by standing down:

```ts
if (isChange && hasTypedSince(inputEl, sel)) return
```

`recordCursor` already returns the `value` it saw, so the check needs no extra bookkeeping.

It is deliberately scoped to the `CHANGE` path. There the write only echoes back what the user
typed, and a newer `INPUT.CHANGE` is guaranteed to have queued its own sync, so skipping is always
safe. The increment, blur and format paths are untouched, and when the input is not focused
`recordCursor` returns `undefined`, so those writes stay unguarded as before.

No.

## 📝 Additional Information

**Measured, before and after.** I patched the guard into the compiled machine, rebuilt the page that
exercises it, then reverted the patch and rebuilt from the same source — same machine, same build,
same 30-iteration loop, only `syncInputElement` different:

| build | failures |
| --- | --- |
| unpatched | **9 / 30** (eight `1.5` → `55`, two `50000` → `5,000`) |
| patched | **0 / 30** |

**Tests.** `hasTypedSince` is covered in `packages/machines/number-input/tests/cursor.test.ts`
(27 passed). Worth flagging: `test-js` globs `packages/{utilities,core}`, so nothing under
`packages/machines` runs in CI today — that file was already there and already unreached. Happy to
add an e2e case instead if you'd prefer, though I'd want to check it actually reproduces through
Playwright's per-key round trip before claiming it guards anything.

`pnpm --filter @zag-js/number-input typecheck` and `lint` are both clean.

No before/after recordings: the failure is a sub-frame race that a video wouldn't show more clearly
than the trace above, and I did not have recording tooling in this session.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

<h2><a href="https://app.greptile.com/api/retrigger?id=66075862"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 4/5</h2>

This PR is not safe to merge until the guard distinguishes a genuinely newer input event from framework-driven formatting writes.

<h2><a href="https://app.greptile.com/ide/claude-code?prompt=%23%23%23%20Issue%201%0Apackages%2Fmachines%2Fnumber-input%2Fsrc%2Fnumber-input.machine.ts%3A451%0AIn%20Vue%2C%20Svelte%2C%20and%20Solid%2C%20%60defaultValue%60%20becomes%20a%20live%20%60value%60%20prop.%20With%20formatting%20enabled%2C%20inserting%20%601%60%20before%20%60999%60%20can%20cause%20the%20renderer%20to%20replace%20the%20recorded%20raw%20value%20%601999%60%20with%20%601%2C999%60%20before%20this%20RAF%20runs%2C%20without%20emitting%20another%20%60INPUT.CHANGE%60.%20The%20guard%20treats%20that%20write%20as%20a%20newer%20keystroke%20and%20skips%20both%20the%20value%20sync%20and%20cursor%20restoration%2C%20so%20the%20next%20character%20can%20be%20inserted%20at%20the%20wrong%20position.%0A%0A%23%23%23%20Issue%202%0Apackages%2Fmachines%2Fnumber-input%2Ftests%2Fcursor.test.ts%3A132%0AThis%20suite%20is%20under%20%60packages%2Fmachines%60%2C%20which%20is%20not%20selected%20by%20the%20unit-test%20targets%20run%20through%20CI's%20%60pnpm%20test%60.%20It%20also%20invokes%20%60hasTypedSince%60%20directly%20instead%20of%20exercising%20the%20deferred%20%60syncInputElement%60%20behavior.%20CI%20can%20therefore%20pass%20if%20this%20race%20fix%20regresses%20or%20if%20non-keystroke%20DOM%20changes%20are%20mishandled%3B%20an%20executed%20machine%20or%20browser%20test%20should%20cover%20multiple%20updates%20before%20the%20RAF.%0A%0ANote%3A%20If%20this%20suggestion%20doesn't%20match%20your%20team's%20coding%20style%2C%20reply%20to%20this%20and%20let%20me%20know.%20I'll%20remember%20it%20for%20next%20time!%0A%0A---%0A%0AFor%20each%20issue%20above%2C%20determine%20whether%20it%20is%20valid%20and%20should%20be%20fixed.%20If%20so%2C%20fix%20it%20directly.&repo=chakra-ui%2Fzag&pr=3352&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaudeDark.svg?v=7"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaude.svg?v=7">Fix All in Claude Code</picture></a>Findings</h2>

1. P1&nbsp;**Renderer writes mimic typing** <a href="https://github.com/chakra-ui/zag/pull/3352#discussion_r4047929600">▶</a>
2. P2&nbsp;**Regression test does not run** <a href="https://github.com/chakra-ui/zag/pull/3352#discussion_r4047929611">▶</a>

<details><summary>Fix with agent prompt</summary>

`````markdown
### Issue 1
packages/machines/number-input/src/number-input.machine.ts:451
In Vue, Svelte, and Solid, `defaultValue` becomes a live `value` prop. With formatting enabled, inserting `1` before `999` can cause the renderer to replace the recorded raw value `1999` with `1,999` before this RAF runs, without emitting another `INPUT.CHANGE`. The guard treats that write as a newer keystroke and skips both the value sync and cursor restoration, so the next character can be inserted at the wrong position.

### Issue 2
packages/machines/number-input/tests/cursor.test.ts:132
This suite is under `packages/machines`, which is not selected by the unit-test targets run through CI's `pnpm test`. It also invokes `hasTypedSince` directly instead of exercising the deferred `syncInputElement` behavior. CI can therefore pass if this race fix regresses or if non-keystroke DOM changes are mishandled; an executed machine or browser test should cover multiple updates before the RAF.

Note: If this suggestion doesn't match your team's coding style, reply to this and let me know. I'll remember it for next time!

---

For each issue above, determine whether it is valid and should be fixed. If so, fix it directly.
`````

</details>

<details><summary>Summary</summary>

This PR adds a DOM-value comparison before number input's deferred CHANGE synchronization, intending to prevent an older RAF callback from overwriting a newer keystroke.

- Adds `hasTypedSince` and uses it to skip stale value and cursor writes.
- Adds focused unit cases for the comparison helper and a patch changeset.
- The comparison also classifies adapter-driven formatting as typing, which can skip the only cursor restoration in Vue, Svelte, and Solid.
- The new unit suite is not exercised by the repository's CI test command and does not reproduce the deferred race.
</details>

<details open><summary>Diagram</summary>

```mermaid
sequenceDiagram
  participant User
  participant DOM as Input DOM
  participant Machine as Number-input machine
  participant Renderer as Vue/Svelte/Solid renderer
  participant RAF as Queued RAF
  User->>DOM: Insert "1" before "999"
  DOM->>Machine: INPUT.CHANGE ("1999", caret 1)
  Machine->>RAF: Queue raw sync and cursor restoration
  Machine-->>Renderer: formattedValue becomes "1,999"
  Renderer->>DOM: Set value to "1,999"
  Note over DOM,Renderer: Programmatic write emits no INPUT.CHANGE
  RAF->>DOM: Compare "1,999" with recorded "1999"
  RAF-->>DOM: Return without restoring value or caret
  User->>DOM: Next key lands at the moved caret
```
</details>

<sub>Reviews (1) · Last reviewed commit: ["fix(number-input): don&#39;t overwrite a key..."](https://github.com/chakra-ui/zag/commit/7e2b8b6d81027439e24ea544af6f7c3c277e9b0a)</sub>


strongpauly:
Both findings are valid. Pushed b8bb50b.

**P1 — renderer writes mimic typing.** Agreed, and it would have been a real regression. Confirmed the chain:

- `getInputProps` binds `defaultValue: computed("formattedValue")`, and `formattedValue` is `formatValue(parseValue(context.value))` — so mid-entry the machine's value is `1999` while the bound prop is `1,999`
- `normalize-props.ts` maps `defaultValue: "value"` in **solid, vue, svelte and vanilla**; react and preact keep `defaultValue`

So on those four the renderer writes the formatted text between the event and the frame, my DOM comparison read it as a keystroke, and skipped the cursor restore — reintroducing #2712 on four of six adapters. I only measured react, via Chakra, which is exactly the blind spot the comment names.

The guard now asks the machine rather than the DOM:

```ts
if (isChange && context.get("value") !== value) return
```

A newer keystroke moves `context.value` — that is what queues the next sync — whereas a render does not touch it. Since it never reads the input, the `defaultValue`/`value` difference cannot reach it.

One thing worth flagging while I was checking that this reads live state inside the `raf`. It does on react (`ctx.get` → `bindable.ref.current`) and on vue/solid/svelte/vanilla (`ctx.get` → `bindable.get()`, a closure variable that `set` mutates). **Preact is the exception**: its `ctx.get` calls `bindable.get()`, which closes over the render's `value`, where react's `ctx.get` deliberately goes through `.ref.current`. So on preact the guard simply never fires — today's behaviour, unchanged, rather than a regression. Aligning preact's `ctx.get` with react's looks like a genuine one-line fix but it is a different package and a different bug, so I have left it out of this PR rather than widen it. Happy to raise it separately if you agree.

**P2 — regression test does not run.** Agreed, and I had flagged the glob in the PR description. I have removed those tests rather than leave them: with the guard no longer built on an extracted predicate there is nothing honest left for them to assert, and `cursor.ts` is back to untouched. The diff is now one file plus the changeset.

I do not have a good replacement, and would rather say so than add a green test that proves nothing:

- a `packages/machines` unit test does not run under `pnpm test`
- an e2e test does run, but reproducing the race needs two keystrokes inside one animation frame, and Playwright's per-key CDP round trip is slower than a frame — `pressSequentially` with no delay already passes on the unfixed code
- dispatching two `input` events synchronously from `page.evaluate` does not reproduce it either: both syncs then land in the same frame and the second one corrects the first, which is precisely the case that was never broken

What did give a signal was driving a real Chromium through `@storybook/test-runner`, where in-page dispatch puts keystrokes ~15 ms apart. Against a bare `NumberInput.Root` — no wrapper, no form library, no cursor fed back by the harness — that reproduced at 9/30, and 0/30 with the fix compiled in. If you would like that shape of test in the repo I am happy to add it, but it needs a runner that dispatches in-page rather than over CDP, so it is a bigger conversation than this PR.

Note the 9/30 → 0/30 figure was measured against the **first** version of the guard. The revision is a strictly narrower condition — `context.value` differing is what schedules the second sync in that exact sequence — but I have not re-run the loop against b8bb50b, and I would rather label that clearly than imply a measurement I have not repeated.

